PR #52, opened 2019-04-01
Change versioning to git

brittonsmith:
This changes the functionality that reports the version of the code being built from hg to git. For now, I've gotten rid of using the `create_config_info.py` Python script since we can do just about everything with bash commands. Instead, we just pipe the output of `WriteConfigure` to the `Enzo_Build` file. This will eliminate the need to have a version of Python with special packages installed in order to get the version information compiled in with the code.

I've also switched the `.hgignore` to a `.gitignore`, removed `.hgtags`, and fixed a couple file permissions that always bugged me.


src/enzo/Makefile
```diff
@@ -143,8 +141,8 @@ MACHNOTES:
 	@(if [ -e $@ ] ; then  \
 		echo "Success!" ; \
 		echo "Compiled enzo from " ; \
-		echo "Mercurial Branch   `$(SVN) identify -b`" ; \
-		echo "Mercurial Revision `$(SVN) identify -i`" ; \
+		echo "Mercurial Branch   `git rev-parse --abbrev-ref HEAD`" ; \
```
chummels: Maybe change: "Mercurial" -> "Git" here?

brittonsmith: Good catch! I've fixed this in PR #55.

jwise77:
Three approvals. Merging. First Enzo PR on github!

PR #55, opened 2019-04-09
Add automated testing with circle-ci

brittonsmith:
This adds automated testing using the circle-ci service. The testing is the same push suite as before with an added test of building the docs as well. We now cache external dependencies, like hypre, and the answers from the gold standard. The latter cuts the test time in half.

I had to disable some of the mercurial version checking in the test runner to get things working. I can eventually replace this with the git counterpart, but I think that can wait until after this is accept. None of that affects the running of the tests.


.circleci/config.yml
```diff
@@ -0,0 +1,173 @@
...
+          flags: --answer-store
+
+      - save_cache:
+          name: "Save test answers cache."
+          key: test-answers-v1
+          paths:
+            - ~/enzo_test/push_suite
+
+      - build-and-test:
+          tag: tip
```
brittonsmith: After this PR is accepted, this line should be changed to:
```
tag: gold-standard-v12
```
And then the appropriate tag added.

brittonsmith: @gregbryan, good question. CircleCI uses an immutable object store for caches, so if we want to change something or use updated dependencies, we have to increment the version number ourselves. Currently, only the `hypre` libraries are getting cached since we install everything else with `apt-get`.

brittonsmith: This system is good, but saying everything out loud makes me realize how many little pieces of information there are to remember. I'm going to make an issue about adding something to the developer documentation.

brittonsmith:
@jwise77, @gregbryan, want to go ahead and merge this in ?

jwise77:
> @jwise77, @gregbryan, want to go ahead and merge this in ?

I was waiting on a third review.  But I think this is an important addition before moving forward, so I'm merging now.

PR #57, opened 2019-04-17
Active particles

fearmayo:
Huge PR for the active particles machinery. The bulk of the original Active Particle machinery was developed by Nathan et al. I've tested it pretty robustly and it works well as far as I see. 
There are some decisions to be made around the IO but at this stage the code is at least ready for review. I will add documentation and some additional tests in due course.


brittonsmith:
@fearmayo, when you have a chance, could you merge this with the latest changes from the master branch? We've now got the test suite running there, so it would be good to see if this PR is passing all those tests.

src/enzo/phys_constants.h
```diff
@@ -70,6 +70,16 @@
 
 #define SolarMass                       1.9891e33
 
+/* Solar Radius [cm] */
+
+#define SolarRadius                     6.96e10
+
+/* Solar Luminosity [erg/s] */
+
+#define SolarLuminosity                 3.9e33
```
ibutsky: I see that SolarLuminosity is only used in a print statement conversion right now, but it would be good to change the definition of SolarLuminosity to 3.839e33 for higher accuracy in future use cases.

src/enzo/hydro_rk/EOS.h
```diff
@@ -16,6 +16,7 @@ inline void EOS(float &p, float &rho, float &e, float &h, float &cs, float &dpdr
        6: minimum pressure similar (but not equal to)
           http://adsabs.harvard.edu/abs/2004ApJ...606...32R
           equation 4
+       7: Burkert & Bodenheimer Test. Fedderath et al. (2010) Eqn
```
ibutsky: Small typo: should be Federrath et al. 2010, Eqn 30-31. Alternatively, can add the link for clarity https://ui.adsabs.harvard.edu/abs/2010ApJ...713..269F

src/enzo/hydro_rk/EOS.h
```diff
@@ -143,6 +144,38 @@ inline void EOS(float &p, float &rho, float &e, float &h, float &cs, float &dpdr
...
+    else if(rho/rho_cr <= 5.0) {
+      new_gamma = 1.1;
+      cs = c_s*pow(rho/rho_cr, new_gamma);
+    }
+    else {
+      new_gamma = 4.0/3.0;
+      cs = c_s*pow(rho/rho_cr, new_gamma);
+    }
+    
+    p = rho*cs*cs;
```
ibutsky: Following equation 30 in Federrath et al. 2010, should this be p = rho* c_s * cs instead?

src/enzo/hydro_rk/EOS.h
```diff
@@ -143,6 +144,38 @@ inline void EOS(float &p, float &rho, float &e, float &h, float &cs, float &dpdr
 
   }
 
+  /* Only works in mode 1 and only updates internal energy */
```
ibutsky: Need to add to the EOS documentation to specify that this only works in mode 1 (or add an assertion in the code)

src/enzo/bondi_alpha.C
```diff
@@ -0,0 +1,56 @@
+#include "preincludes.h"
+#include "macros_and_parameters.h"
+
+
+/* Routine to return alpha, defined as rho/rho_inf, for a critical
+   Bondi accretion solution.  The argument is x = r / rBondiHoyle. 
+   Adapted from Orion, courtesy of Mark Krumholz */
```
ibutsky: Does this follow the approach described in Krumholz et al. 2004(https://ui.adsabs.harvard.edu/abs/2004ApJ...611..399K)? It would be helpful to include a link to a description of the equations if it's easily available.

src/enzo/RecordTotalActiveParticleCount.C
```diff
@@ -0,0 +1,62 @@
+/***********************************************************************
+/
+/  RECORD TOTAL NUMBER OF ACTIVE PARTICLES
+/
+/  written by: Ji-hoon Kim
+/  date:       October, 2009
+/  modified1:  December, 2011 by John Wise -- active particle counts
+/
+/  PURPOSE:
```
ibutsky: Update purpose/written by history (if applicable)

src/enzo/ActiveParticle.h
```diff
@@ -0,0 +1,816 @@
...
+
+      for (AttributeVector::iterator it = handlers.begin();
+          it != handlers.end(); ++it) {
+	  const char *_name = (*it)->name.c_str();
+	  /* HACK - there must be a better way to do this...*/
+	  if( (strcmp(_name, "AccretionRate") == 0 ||
+	       strcmp(_name, "AccretionRateTime") == 0 ||
+	       strcmp(_name, "Accreted_angmom") == 0)
+	      && strcmp(_ap_name, "SmartStar") == 0)
+	    {
```
DeovratPrasad: This looks very problem specific hack and a more generalized way of doing this must be implemented before merging.

src/enzo/RecordTotalActiveParticleCount.C
```diff
@@ -0,0 +1,62 @@
...
+#include "GridList.h"
+#include "ExternalBoundary.h"
+#include "Grid.h"
+#include "Hierarchy.h"
+#include "TopGridData.h"
+#include "LevelHierarchy.h"
+#include "CosmologyParameters.h"
+#include "CommunicationUtilities.h"
+
+void RecordTotalActiveParticleCount(HierarchyEntry *Grids[], int NumberOfGrids,
```
ibutsky: This method is declared in ActiveParticleInitialize, but isn't used. Is this functionality duplicated elsewhere?

src/enzo/RecordTotalActiveParticleCount.C
```diff
@@ -0,0 +1,62 @@
...
+  }
+ 
+#ifdef USE_MPI
+  /* Get counts from each processor to get total list of new particles. */
+ 
+  MPI_Datatype DataTypeInt = (sizeof(int) == 4) ? MPI_INT : MPI_LONG_LONG_INT;
+
+  MPI_Arg GridCount = NumberOfGrids;
+   
+  MPI_Allreduce(PartialActiveParticleCountPrevious,
```
ibutsky: If USE_MPI isn't defined, it looks like TotalActiveParticleCountPrevious doesn't get updated

src/enzo/ActiveParticle.h
```diff
@@ -0,0 +1,816 @@
+/*-*-C++-*-*/
+/***********************************************************************
+/
+/  STAR PARTICLE STRUCTURE
+/
+/  written by: John Wise
+/  date:       September, 2005
+/  modified1:  John Wise
+/  date:       March, 2009 (converted into a class)
```
DeovratPrasad: Update it to Active Particle Structure

src/enzo/RebuildHierarchy.C
```diff
@@ -395,6 +402,19 @@ int RebuildHierarchy(TopGridData *MetaData,
       tt1 = ReturnWallTime();
       RHperf[3] += tt1-tt0;
 
+      /* Deposit flagging field for active particles.
+
+     NJG:  This is a hack so that the accretion zone of sink particles are
+           resolved at all times.  While this will work fine for small number of
+           particles, there will be scaling issues if one tries to do nonlocal
+           refinement on large numbers of particles
```
ibutsky: After how many sink particles do the scaling issues become important? This would be good to include as a warning in the documentation.

src/enzo/ActiveParticleFinalize.C
```diff
@@ -0,0 +1,94 @@
+/***********************************************************************
+/
+/  ACTIVE PARTICLE INITIALIZATION
+/
+/  written by: John Wise
+/  date:       March, 2009
+/  modified1:  November, 2011 (JHW) -- converting to active particles
+/
+/  PURPOSE: Contains all routines to finalize the star particles.
+/
```
DeovratPrasad: Update the text for purpose of this file.

src/enzo/ActiveParticleFinalize.C
```diff
@@ -0,0 +1,94 @@
...
+  int i;
+
+  if (EnabledActiveParticlesCount == 0) return SUCCESS;
+  FLOAT TimeNow = LevelArray[level]->GridData->ReturnTime();
+  float Timestep = LevelArray[level]->GridData->ReturnTimeStep();
+
+  LCAPERF_START("ActiveParticleFinalize");
+
+  /* Update the star particle counters. */
+
```
DeovratPrasad: Change star particle to Active Particle

src/enzo/ActiveParticleInitialize.C
```diff
@@ -0,0 +1,119 @@
+/***********************************************************************
+/
+/  ACTIVE PARTICLE INITIALIZATION
+/
+/  written by: John Wise
+/  date:       March, 2009
+/  modified1:  November, 2011 (JHW) -- Converted to active particles.
+/
+/  PURPOSE: Contains all routines to initialize the star particles.
+/
```
DeovratPrasad: Update the purpose of this file.

src/enzo/ActiveParticleRoutines.C
```diff
@@ -0,0 +1,508 @@
...
+/
+/  ROUTINES FOR THE STAR PARTICLE CLASS
+/
+/  written by: John Wise
+/  date:       March, 2009
+/  modified1:  November, 2011 (converting into active particles)
+/
+/  PURPOSE: Instead of restricting star particles to the typical 
+/           particle attributes, this class gives more functionality 
+/           to them.
```
DeovratPrasad: Update the documentation about the purpose of this file.

src/enzo/ActiveParticleRoutines.C
```diff
@@ -0,0 +1,508 @@
...
+#define FACTORINCREASE 1e10
+#define MAX_ENERGY 1e7
+#define DENSITY_WEIGHTED 1
+#define RAMPTIME 1e4
+/* We need to make sure that we can operate on the grid, so this dance is
+ * necessary to make sure that grid is 'friend' to this particle type. */
+
+class SmartStarGrid : private grid {
+  friend class ActiveParticleType_SmartStar;
+};
```
DeovratPrasad: Add comment about the purpose of this class.

src/enzo/ActiveParticleRoutines.C
```diff
@@ -0,0 +1,508 @@
...
+     in ActiveParticleFinalize.*/
+  Identifier = INT_UNDEFINED;
+
+  // Assume newly created particles will not be immediately deleted
+  WillDelete = false;
+}
+
+
+ActiveParticleType::~ActiveParticleType(void)
+{
```
DeovratPrasad: Is this function needed?

src/enzo/ActiveParticleRoutines.C
```diff
@@ -0,0 +1,508 @@
...
+  float DensityUnits, LengthUnits, TemperatureUnits, TimeUnits,
+    VelocityUnits, MassConversion;
+  GetUnits(&DensityUnits, &LengthUnits, &TemperatureUnits,
+	   &TimeUnits, &VelocityUnits, CurrentGrid->Time);
+  dx = LengthUnits * CurrentGrid->CellWidth[0][0];
+  MassConversion = (float) (dx*dx*dx * double(DensityUnits) / Msun);
+  this->Mass *= MassConversion;
+  return;
+}
+
```
DeovratPrasad: I think function 'ConvertAllMassesToSolar' and 'ConvertMassToSolar' are identical. Is both required?

src/enzo/New_Grid_ReadGrid.C
```diff
@@ -194,8 +197,18 @@ int grid::Group_ReadGrid(FILE *fptr, int GridID, HDF5_hid_t file_id,
...
     }
- 
-    if (NumberOfParticles > 0) {
+
+    if (fscanf(fptr, "NumberOfActiveParticles = %"ISYM"\n", &NumberOfActiveParticles) != 1) {
+      // Ugly hack to support restart files from before active particle support
+      NumberOfActiveParticles = 0;
+    }
+
+    // We read these in but don't use them (yet?). Therefore we won't error
```
ibutsky: Are there still plans to use this?

src/enzo/ActiveParticleRoutines.C
```diff
@@ -0,0 +1,508 @@
...
+    vel[dim] = ratio1 * vel[dim] + ratio2 * a->vel[dim];
+    //accreted_angmom[dim] = ratio1 * accreted_angmom[dim] + ratio2 * a.accreted_angmom[dim];
+  }
+  Mass += a->Mass;
+  //FinalMass += a.FinalMass;
+  //DeltaMass += a.DeltaMass;
+  //last_accretion_rate += a.last_accretion_rate;
+  //NotEjectedMass += a.NotEjectedMass;
+  return;
+}
```
DeovratPrasad: Many commented out lines in this function. Slight clean will make it look better.

src/enzo/ActiveParticleRoutines.C
```diff
@@ -0,0 +1,508 @@
...
+#ifdef UNUSED
+bool ActiveParticleType::MergableMBH(ActiveParticleType *a)
+{
+  // Merge MBH particle with another 
+  return type == a->type && type == MBH;
+}
+#endif
+
+float ActiveParticleType::Separation2(ActiveParticleType *a)
+{
```
DeovratPrasad: Add a comment about the purpose of this function.

src/enzo/ActiveParticleRoutines.C
```diff
@@ -0,0 +1,508 @@
...
+  int dim;
+  float dr, result = 0;
+  for (dim = 0; dim < MAX_DIMENSION; dim++) {
+    dr = pos[dim] - a->pos[dim];
+    result += dr*dr;
+  }
+  return result;
+}
+
+float ActiveParticleType::Separation(ActiveParticleType *a)  { return sqrt(this->Separation2(a)); }
```
DeovratPrasad: What is the purpose of this function?

src/enzo/ActiveParticleRoutines.C
```diff
@@ -0,0 +1,508 @@
...
+  for (dim = 0; dim < MAX_DIMENSION; dim++) {
+    dv = vel[dim] - a->vel[dim];
+    result += dv*dv;
+  }
+  return result;
+}
+
+void ActiveParticleType::UpdatePositionVelocity(void)
+{
+  LCAPERF_START("star_UpdatePositionVelocity");
```
DeovratPrasad: Is this 'star' or 'Active particle' position velocity function?

src/enzo/ActiveParticleRoutines.C
```diff
@@ -0,0 +1,508 @@
...
+      vel[dim] = CurrentGrid->ParticleVelocity[dim][_id];
+    }
+  }
+  LCAPERF_STOP("star_UpdatePositionVelocity");
+  return;
+}
+
+void ActiveParticleType::CopyFromParticle(grid *_grid, int _id, int _level)
+{
+  int dim;
```
DeovratPrasad: Add comment about the purpose of this function.

src/enzo/ActiveParticleRoutines.C
```diff
@@ -0,0 +1,508 @@
...
+
+  // No more attributes.  Everything stored in active particles.
+//  BirthTime = _grid->ParticleAttribute[0][_id];
+//  DynamicalTime = _grid->ParticleAttribute[1][_id];
+//  Metallicity = _grid->ParticleAttribute[2][_id];
+
+  // below is removed because we want to keep Star->Mass as double 
+  // during the run - Ji-hoon Kim, Dec.2009
+//  Mass = (double)(_grid->ParticleMass[_id]); 
+//  this->ConvertMassToSolar();
```
DeovratPrasad: These commented out lines needs to be purged.

src/enzo/ActiveParticleRoutines.C
```diff
@@ -0,0 +1,508 @@
...
+      source->Position[dim] = this->pos[dim];
+  }
+  source->IsActiveParticle = true;
+  return source;
+}
+#endif /* TRANSFER */
+
+int ActiveParticleType_SmartStar::CalculateAccretedAngularMomentum()
+{
+
```
DeovratPrasad: Need to add comment about the use of this function.

src/enzo/ActiveParticle_Kravtsov.C
```diff
@@ -0,0 +1,220 @@
...
+  // cc and (d1/mproton) gives the mean density of the universe in
+  // particles/cm^3 (assuming hydrogen is dominant)
+
+  densthresh = DensityThreshold / (supp_data.DensityUnits / mh);
+
+
+  // calculate time constant for star formation.  This assumes that
+  // the user input is in units of years
+  
+  timeconstant = StarFormationTimeConstant * 3.156e7 / supp_data.TimeUnits;
```
jwise77: You can use the `yr_s` define from `phys_const.def` for consistency.

src/enzo/ActiveParticle_GalaxyParticle.C
```diff
@@ -0,0 +1,445 @@
...
+			   int NumberOfGrids, int SendField);
+
+int ActiveParticleType_GalaxyParticle::SubtractMassFromGrid(int nParticles,
+    ActiveParticleList<ActiveParticleType>& ParticleList, LevelHierarchyEntry *LevelArray[],
+    FLOAT dx, int ThisLevel)
+{
+
+  // This function is a work in progress.
+  return SUCCESS;
+
```
DeovratPrasad: This function has all the arguments commented out and is not being used. It should be removed.

src/enzo/ActiveParticle_GalaxyParticle.h
```diff
@@ -0,0 +1,207 @@
...
+      
+      if (nParticles == 0)
+	return SUCCESS;
+
+      /* Calculate CellWidth on maximum refinement level */
+
+      // This assumes a cubic box and may not work for simulations with
+      // MinimumMassForRefinementLevelExponent
+      FLOAT dx = (DomainRightEdge[0] - DomainLeftEdge[0]) /
+	(MetaData->TopGridDims[0]*POW(FLOAT(RefineBy),FLOAT(MaximumRefinementLevel)));
```
DeovratPrasad: Has this been tested?

src/enzo/phys_constants.h
jwise77: On this same topic, I've seen a few hard coded physical constants in the new code. There was a recent PR (in [bitbucket](https://bitbucket.org/enzo/enzo-dev/pull-requests/417)) that made all of these changes.

I would be fine with approving the changes without this modification and making an issue to address this later because APs would still function but be less consistent with the precision issues.

src/enzo/ActiveParticle_GalaxyParticle.h
```diff
@@ -0,0 +1,207 @@
...
+	  // Find the period of the box.
+	  // I'm going to be cheap here because galaxy particles are only ever
+	  // meant to be run in 3D.
+	  FLOAT period[3];
+	  for (int dim = 0; dim < 3; dim++) {
+	    period[dim] = DomainRightEdge[dim] - DomainLeftEdge[dim];
```
DeovratPrasad: What does being cheap mean in this context?

src/enzo/ActiveParticle_AccretingParticle.h
```diff
@@ -0,0 +1,345 @@
...
+#include "ExternalBoundary.h"
+#include "Grid.h"
+#include "Hierarchy.h"
+#include "LevelHierarchy.h"
+#include "TopGridData.h"
+#include "CommunicationUtilities.h"
+#include "communication.h"
+#include "phys_constants.h"
+#include "FofLib.h"
+#define DEBUG 1
```
jwise77: Should be this turned off?

src/enzo/Grid_RemoveMassFromGrid.C
```diff
@@ -0,0 +1,469 @@
...
+#include "GridList.h"
+#include "ExternalBoundary.h"
+#include "fortran.def"
+#include "Grid.h"
+#include "phys_constants.h"
+#include "ActiveParticle_SmartStar.h"
+#define TINY_NUMBER         1e-20
+#define SMALL_NUMBER         1e-6
+#define ACCRETION_LIMIT     9e-1
+#define N 8
```
ibutsky: Should probably be undefined at the end of this file (or moved to macros_and_parameters if they're meant to be global).

src/enzo/ActiveParticle_PopIII.C
```diff
@@ -0,0 +1,263 @@
+
+/***********************************************************************
+/
+/  AN EXAMPLE ACTIVE PARTICLE TYPE
+/
+/  written by: Matthew Turk
+/  date:       May, 2011
+/
+/  PURPOSE:
```
DeovratPrasad: Add the purpose of this file.

src/enzo/ActiveParticle_RadiationParticle.C
```diff
@@ -0,0 +1,524 @@
...
+  // RadiationSED[2] = 100.0/RadiationSEDNumberOfBins;
+  // RadiationEnergyBins[3] = 12.8;
+  // RadiationSED[3] = 100.0/RadiationSEDNumberOfBins;
+  // RadiationEnergyBins[4] = 1;
+  // RadiationSED[4] = 100.0/RadiationSEDNumberOfBins;
+  // RadiationEnergyBins[5] = 200.0;
+  // RadiationSED[5] = 100.0/RadiationSEDNumberOfBins;
+  RadiationLifetime = RadiationSourceLifeTime;
+  RadiationPhotonsPerSecond = PhotonsPerSecond;
+  FixedInSpace = TRUE; //Hardcoded for the moment but could potentially be a parameter
```
DeovratPrasad: It should be made into a parameter before merging.

src/enzo/ActiveParticle_SmartStar.C
```diff
@@ -0,0 +1,1058 @@
...
+	// printf("Thermal Energy = %1.20g\n", thisGrid->BaryonField[data.GENum][index]);
+	// printf("Total - Thermal = %g\n", thisGrid->BaryonField[data.TENum][index] -
+	//        thisGrid->BaryonField[data.GENum][index]);
+	// float vel = pow(velx[index], 2.0) + pow(vely[index], 2.0) + pow(velz[index], 2.0);
+	// printf("Kinetic Energy =  %1.20g\n", 0.5*vel);
+	// printf("Total - Thermal - Kinetic = %g\n", thisGrid->BaryonField[data.TENum][index] -
+	//        thisGrid->BaryonField[data.GENum][index] - 0.5*vel);
+	// printf("Gravitational Energy = %1.20g\n", thisGrid->BaryonField[data.GravPotentialNum][index]);
+	// printf("Potential Field = %1.20g\n", thisGrid->PotentialField[index]);
+	// printf("mass = %g\n", mass);
```
DeovratPrasad: Remove these commented out bunch of print statements.

src/enzo/ActiveParticle_SmartStar.C
```diff
@@ -0,0 +1,1058 @@
...
+
+
+template <class active_particle_class>
+int ActiveParticleType_SmartStar::BeforeEvolveLevel
+(HierarchyEntry *Grids[], TopGridData *MetaData,
+ int NumberOfGrids, LevelHierarchyEntry *LevelArray[], 
+ int ThisLevel, bool CallEvolvePhotons, 
+ int TotalStarParticleCountPrevious[],
+ int SmartStarID)
+{
```
DeovratPrasad: need some info about this function in context of SmartStar

src/enzo/ActiveParticle_SmartStar.h
```diff
@@ -0,0 +1,450 @@
...
+  static int RadiationSEDNumberOfBins;
+  static float* RadiationEnergyBins;
+  static float* RadiationSED;
+  static float RadiationLifetime;
+  //float acc[3];
+  int ParticleClass;
+
+  float AccretionRate[NTIMES];  //In would be nice to make this a vector
+  //std::vector<float> AccretionRateTime;
+  float AccretionRateTime[NTIMES];
```
DeovratPrasad: If it makes things better, these should be made into vectors before merging.

src/enzo/ActiveParticle_SmartStar.h
```diff
@@ -0,0 +1,450 @@
...
+template <class active_particle_class>
+int ActiveParticleType_SmartStar::AfterEvolveLevel(
+    HierarchyEntry *Grids[], TopGridData *MetaData,
+    int NumberOfGrids, LevelHierarchyEntry *LevelArray[], 
+    int ThisLevel, int TotalStarParticleCountPrevious[],
+    int SmartStarID)
+{
+
+  /* Accreting particles live on the maximum refinement level.  If we are on a lower level, this does not concern us */
+
```
DeovratPrasad: Why is 'Accreting particle' being used here?

src/enzo/Grid_FindMinimumPotential.C
```diff
@@ -0,0 +1,108 @@
...
+	pos[0] = CellLeftEdge[0][i] + 0.5*CellWidth[0][i];
+	pos[1] = CellLeftEdge[1][j] + 0.5*CellWidth[1][j];
+	pos[2] = CellLeftEdge[2][k] + 0.5*CellWidth[2][k];
+	
+
+	for (kk = GridStartIndex[2]; kk <= GridEndIndex[2]; kk++) {
+	  for (jj = GridStartIndex[1]; jj <= GridEndIndex[1]; jj++) {
+	    for (ii = GridStartIndex[0]; ii <= GridEndIndex[0]; ii++) {
+	      lindex = GRIDINDEX_NOGHOST(ii, jj, kk);
+	      mass_j = density[lindex]*dx*dx*dx;
```
ibutsky: This will give an inaccurate estimate of the cell volume if the cell dimensions aren't all equal.

src/enzo/Grid_FindMassinRegion.C
```diff
@@ -0,0 +1,71 @@
...
+	index = GRIDINDEX_NOGHOST(i, j, k);
+	pos[0] = CellLeftEdge[0][i] + 0.5*CellWidth[0][i];
+	pos[1] = CellLeftEdge[1][j] + 0.5*CellWidth[1][j];
+	pos[2] = CellLeftEdge[2][k] + 0.5*CellWidth[2][k];
+	/* Use only values within the control volume */
+	if(sqrt((cellpos[0] - pos[0])*(cellpos[0] - pos[0]) +
+		(cellpos[1] - pos[1])*(cellpos[1] - pos[1]) +
+		(cellpos[2] - pos[2])*(cellpos[2] - pos[2])) < radius)
+	{
+	  TotalMass += density[index]*dx*dx*dx;
```
ibutsky: This will give an inaccurate TotalMass if the cell dimensions aren't all equal

src/enzo/ActiveParticle_SpringelHernquist.C
```diff
@@ -0,0 +1,245 @@
...
+
+	// Calculate the fraction of mass in cold clouds.
+	if (y <= 0) {
+	  x = 0;
+	} else {
+	  x = 1. + 1./ 2. / y - sqrt(1. / y + 1. / 4. / y / y); // Eq(18)
+	}
+
+	// Calculate total baryon mass in the cell.
+	bmass = density[index] * supp_data.MassUnits;
```
DeovratPrasad: baryon mass = density*mass unit??????

src/enzo/CommunicationShareActiveParticles.C
```diff
@@ -0,0 +1,185 @@
...
+
+      if (NumberToSend != 0)
+	local_buffer_size = NumberToSend*instance_size;
+      else
+	local_buffer_size = 0;
+
+      int send_buffer_size;
+      char *send_buffer = new char[local_buffer_size];
+      // This will break if more than one AP type is in the simulation
+      send_buffer_size = ap_info->FillBuffer(SendList, NumberToSend, send_buffer);
```
DeovratPrasad: It does break for multi-type AP in the simulation. Needs to be fixed before merging.

src/enzo/CheckOverlap.C
```diff
@@ -0,0 +1,89 @@
+/***********************************************************************
+/
+/  DETERMINE IF TWO BOXES OVERLAP
+/
+/  written by: Stephen Skory
+/  date:       September, 2012
+/  modified1:
+/
+/  PURPOSE:
```
DeovratPrasad: Needs a bit of text regarding the purpose of this file.

src/enzo/Grid_FindAngularMomentumMinimum.C
```diff
@@ -0,0 +1,81 @@
...
+	pos[1] = CellLeftEdge[1][j] + 0.5*CellWidth[1][j];
+	pos[2] = CellLeftEdge[2][k] + 0.5*CellWidth[2][k];
+	/* Use only values within the control volume */
+	if(sqrt((cellpos[0] - pos[0])*(cellpos[0] - pos[0]) +
+		(cellpos[1] - pos[1])*(cellpos[1] - pos[1]) +
+		(cellpos[2] - pos[2])*(cellpos[2] - pos[2])) <= radius){
+	  ;
+	}
+	else {
+	  continue;
```
ibutsky: It would be more concise to express lines 49-56 as: if(sqrt(.....) > radius) { continue;}

src/enzo/Grid_CalculateSpecificQuantities.C
```diff
@@ -0,0 +1,124 @@
...
+#include "macros_and_parameters.h"
+#include "typedefs.h"
+#include "global_data.h"
+#include "Fluxes.h"
+#include "GridList.h"
+#include "ExternalBoundary.h"
+#include "CosmologyParameters.h"
+#include "Grid.h"
+#include "phys_constants.h"
+#define N 8
```
ibutsky: N isn't undefined later

fearmayo:
> @fearmayo, when you have a chance, could you merge this with the latest changes from the master branch? We've now got the test suite running there, so it would be good to see if this PR is passing all those tests.
@brittonsmith Yes - I meant to do this. Is the best thing to do the merge on Github or best to do it on command line? Will the code comments be preserved after a merge?

src/enzo/Grid_ApplyGalaxyParticleGravity.C
```diff
@@ -0,0 +1,117 @@
...
+//   printf("%d %d %d : %d %d %d\n", this->GridStartIndex[0], this->GridStartIndex[1],
+//     this->GridStartIndex[2], this->GridEndIndex[0], this->GridEndIndex[1],
+//     this->GridEndIndex[2]);
+//   printf("%f %f %f\n", this->CellLeftEdge[0][3] + 0.5*this->CellWidth[0][3],
+//     this->CellLeftEdge[1][3] + 0.5*this->CellWidth[1][3],
+//     this->CellLeftEdge[2][3] + 0.5*this->CellWidth[2][3]);
+  
+    for (k = 0; k < GravitatingMassFieldDimension[0]; k++) {
+      for (j = 0; j < GravitatingMassFieldDimension[1]; j++) {
+        index = GRIDINDEX_NOGHOST(0, j, k);
```
ibutsky: Hard-coded 0 instead of GridStartIndex[0]. I'm not sure if this is always equivalent, but might be better to switch to the more general version.

src/enzo/Grid_ActiveParticleHandler.C
```diff
@@ -0,0 +1,249 @@
...
+
+  /* initialize */
+
+  LCAPERF_START("grid_ActiveParticleHandler");
+
+  /* First we identify the data dependencies */
+
+  struct ActiveParticleFormationDataFlags flags = flags_default;
+
+  int i;
```
dcollins4096: This loop seems strange.  Is it actually doing anything?  This is likely a shortcoming on my end, but it feels like all this does is set the values of `flags` and do nothing with it...

src/enzo/DetermineSEDParameters.C
```diff
@@ -0,0 +1,290 @@
...
+    int arrayindex = CalculateArrayIndex(BHMass, accrate);
+    SS->LuminosityPerSolarMass = BHArray[arrayindex][0]/BHMass; //erg/s/msun
+    
+    if(SmartStarSuperEddingtonAdjustment == TRUE) {
+      if(accrate > eddrate) {
+	//printf("%s: Super Eddington Accretion rate detected (%f Medd) - need to modify feedback!!!!!!!!!\n", __FUNCTION__, accrate/eddrate);
+	float mue = 1.22, a = 0.5;
+	double Ledd = 4*M_PI*GravConst*BHMass*SolarMass*mh*mue*clight/sigma_thompson; //cgs
+	double medddot = 16.0*Ledd/(clight*clight); //cgs
+	double accrate_cgs = accrate*SolarMass/3.154e7;
```
ibutsky: Replace the hard-coded year in seconds with "yr_s" from phys_const.def (also appears throughout this file).

src/enzo/Grid_ActiveParticleHandler.C
```diff
@@ -0,0 +1,249 @@
...
+  //if (debug) printf("StarParticle: end\n");
+
+  LCAPERF_STOP("grid_ActiveParticleHandler");
+  return SUCCESS;
+}
+
+
+int grid::ActiveParticleHandler_Convert(HierarchyEntry* SubgridPointer, int level,
+                                int gridnum, int &NumberOfNewParticles)
+{
```
dcollins4096: Does this Convert function get used anywhere?

src/enzo/ConvertParticles2ActiveParticles.C
```diff
@@ -0,0 +1,272 @@
...
+#include "ExternalBoundary.h"
+#include "Grid.h"
+#include "Hierarchy.h"
+#include "TopGridData.h"
+#ifdef TRANSFER
+#include "ImplicitProblemABC.h"
+#endif
+#include "CosmologyParameters.h"
+
+void CollectParticleTypes(char **active_particle_types, int numparticles);
```
ibutsky: CollectParticleTypes and RebuildHierarchy aren't used below

src/enzo/ActiveParticle_SpringelHernquist.C
```diff
@@ -0,0 +1,245 @@
...
+	// 1. Finest level of refinement
+	if (tg->BaryonField[tg->NumberOfBaryonFields][index] != 0.0) 
+	  continue;
+	
+	// 2. Density greater than threshold
+	if (density[index] < OverDensityThreshold)
+	  continue;
+
+	// Calculate star formation timescale. Eq 21.
+	tstar =  31556926.0 * MinimumDynamicalTime *
```
ibutsky: Replace hard-coded value to yr_s from phys_const.def

src/enzo/DetermineSEDParameters.C
```diff
@@ -0,0 +1,290 @@
...
+  
+  /*
+   * The BH luminosity depends on the Black Hole mass and the accretion rate
+   * The values are calculated assuming a multi-colour disc for the 
+   * for accretion disk and a corona fit by a power law.
+   * The values are hardcoded into the table above and valid for black
+   * hole masses from 1e0 to 1e9 and accretion rates from 1e-7 to 1e2
+   */
+  else if(SS->ParticleClass == BH && SmartStarBHRadiativeFeedback == TRUE) {
+    SS->RadiationLifetime = 1e14*3.154e7/TimeUnits; //code time
```
DeovratPrasad: Year in seconds is hard coded.  There are multiple instances in this file where it has been done. Please, define yr_s in phys_constants.h and use here.

src/enzo/ActiveParticle_SmartStar.C
```diff
@@ -0,0 +1,1058 @@
...
+    if(BH == ThisParticle->ParticleClass)
+      continue;
+    /* The first form of feedback is accretion luminosity 
+     * This is given by L = G M_acc M_* / R_*
+     */
+    /* StarMass in units of Msolar */
+    float StarMass = ThisParticle->Mass * MassConversion / SolarMass;
+    /* Accretion Rate in units of Msolar/s */
+    float AccretionRate = ThisParticle->AccretionRate[ThisParticle->TimeIndex]*MassUnits/(SolarMass*TimeUnits);
+    if(AccretionRate*3.154e7 < 1e-30) /* Takes care of any negative accretion rates */
```
ibutsky: Replace hard-coded 3.154e7 throughout the file with yr_s from phys_const.def

src/enzo/ActiveParticle_SmartStar.C
```diff
@@ -0,0 +1,1058 @@
...
+      printf("%s: Radiation Lifetime =  %e yrs\n", __FUNCTION__, ThisParticle->RadiationLifetime*TimeUnits/3.154e7);
+      printf("%s: Particle Class = %d\t POPIII = %d\n", __FUNCTION__, ThisParticle->ParticleClass, POPIII);
+#endif
+      /* Check for star death and transition to BH */
+      if(ThisParticle->RadiationLifetime < Age) {/* Star needs to go supernovae and change type */
+#ifdef SNEFEEDBACK
+	/* Based on the stellar mass calculate the SN energy and ejecta mass */
+	if(POPIII == ThisParticle->ParticleClass) {
+	  float he_core = (13.0/24.0) * (StarMass - 20.0);
+	  sn_nrg_thistimestep = (5.0 + 1.304 * (he_core - 64.0)) * 1e51; //ergs
```
ibutsky: Is the supernova energy always 1e51? Or is this a parameter that can be specified by the user?

src/enzo/DetermineSEDParameters.C
```diff
@@ -0,0 +1,290 @@
...
+   * temperature of 6000K. 
+   * Luminosity is NOT in ergs/s but in photons/s 
+   */
+  else if (SS->ParticleClass == SMS) {
+    SS->RadiationLifetime = SmartStarSMSLifetime*3.154e7/TimeUnits; //Code Times
+    SS->LuminosityPerSolarMass = 1.4e51/500.0;
+    /* Ideally we call SLUG here. Hardcoded for now */
+    for(int bin = 0; bin < NUMRADIATIONBINS; bin++) {
+      SS->RadiationEnergyBins[bin] = SMSEnergyBins[bin];
+      SS->RadiationSED[bin] = SMSSEDFracBins[bin];
```
DeovratPrasad: Call SLUG before merging if possible.

src/enzo/ActiveParticle_Skeleton.C
```diff
@@ -0,0 +1,309 @@
...
+  {
+    for (j = thisGrid->GridStartIndex[1]; j <= thisGrid->GridEndIndex[1]; j++)
+    {
+      index = GRIDINDEX_NOGHOST(thisGrid->GridStartIndex[0], j, k);
+      for (i = thisGrid->GridStartIndex[0]; i <= thisGrid->GridEndIndex[0]; i++, index++)
+      {
+
+        // Physics goes here.
+
+        if (false)
```
ibutsky: Is the block of code after "if(false)" still necessary?

src/enzo/Grid_AccreteOntoAccretingParticle.C
```diff
@@ -0,0 +1,477 @@
...
+	    if (HydroMethod == PPM_DirectEuler)
+	    {
+	      BaryonField[Vel1Num][index] = pnew[0][index]/mnew[index];
+	      BaryonField[Vel2Num][index] = pnew[1][index]/mnew[index];
+	      BaryonField[Vel3Num][index] = pnew[2][index]/mnew[index];
+	    }
+	    else if (HydroMethod == Zeus_Hydro)
+	    {
+	      // do nothing for now.
+	    }
```
DeovratPrasad: Would be useful to  add the upgrade of the BaryonField for Zeus_Hydro method before merging.

src/enzo/New_Grid_WriteGrid.C
```diff
@@ -808,7 +835,34 @@ int grid::Group_WriteGrid(FILE *fptr, char *base_name, int grid_id, HDF5_hid_t f
...
   } // end: if (NumberOfParticles > 0)
- 
+
+    /* ------------------------------------------------------------------- */
+  /* 4) Save active particle quantities. */
+
+  if (NumberOfActiveParticles > 0) {
+    /* Iterate over the enabled active particle types */
+
+    ParticleGroupID = H5Gcreate(group_id, "Active Particles", 0);
```
DeovratPrasad: Reminder - Need to make YT read the 'Active Particles' data.

src/enzo/ActiveParticle.C
```diff
@@ -0,0 +1,411 @@
...
+     generate them.
+
+     This uses code from the old grid::StarParticleHandler routine. */
+
+
+  /* initialize */
+
+  int dim, i, j, k, index, size, field, GhostZones = NumberOfGhostZones;
+  int DensNum, GENum, TENum, Vel1Num, Vel2Num, Vel3Num, B1Num, B2Num, B3Num,H2INum, H2IINum;
+  const double m_h = 1.673e-24;
```
ibutsky: This is already defined as "mass_h" in phys_const.def

src/enzo/ActiveParticleFindAll.C
```diff
@@ -0,0 +1,258 @@
...
+	    for (i = 0; i < LocalNumberOfActiveParticles; i++)
+	      LocalActiveParticlesOfThisType.copy_and_insert(
+              *ParticlesOnThisProc[i]);
+	    ParticlesOnThisProc.clear();
+	  }
+
+	  /* Finally, append the new active particles to the list */
+	  for(i = LocalNumberOfActiveParticles; i < offset + LocalNumberOfActiveParticles; i++) 
+	    LocalActiveParticlesOfThisType.copy_and_insert(
+            *LocalActiveParticlesOnThisLevel[i - LocalNumberOfActiveParticles]);
```
DeovratPrasad: How is the Active Particle type accounted here?

src/enzo/phys_constants.h
ibutsky: I'm fine with that! I made similar comments throughout, but I understand that these are minor issues that can be addressed later (if they haven't already been addressed in the bitbucket PR).

src/enzo/FofLib.C
```diff
@@ -0,0 +1,1477 @@
...
+  /* Free up unneeded memory */
+  *groupsize=(int *) realloc(*groupsize, groupnum*sizeof(int));
+  *grouplist=(int **) realloc(*grouplist, groupnum*sizeof(int *));
+  free(idxlist);
+  free(root+1);
+
+  return(groupnum); /* Return number of groups */
+}
+
+int Fof(int npart, FLOAT *x, FLOAT link, int *group, int **groupsize)
```
ibutsky: Is this function used?

src/enzo/ActiveParticle_SmartStar.C
jwise77: This is a fit from Heger & Woosley (2002), and I believe it shouldn't be user specified.

src/enzo/phys_constants.h
fearmayo: I think this comment is outside the scope of this PR. Someone should probably update this value in future though.

src/enzo/hydro_rk/EOS.h
fearmayo: Done

src/enzo/bondi_alpha.C
fearmayo: Yes exactly. I've included a link in the file now.

src/enzo/ActiveParticle.h
fearmayo: Agreed! Lets come back to this.

src/enzo/RecordTotalActiveParticleCount.C
fearmayo: There is a function prototype in ActiveParticleInitialize.C alright. I suspect someone used it at some point there. I've now removed the prototype and it looks like right now the code is not called from anywhere. I think it's best to just leave this code here. It may be useful in the future.

src/enzo/RecordTotalActiveParticleCount.C
fearmayo: It gets defined in the build process when MPI is used.

src/enzo/RebuildHierarchy.C
fearmayo: Not sure. This could be difficult to determine and would also depend on the specific problem. This may be a good point to raise at the meeting or on the slack channel.

src/enzo/ActiveParticleFinalize.C
fearmayo: Done

src/enzo/ActiveParticleFinalize.C
fearmayo: Done

src/enzo/ActiveParticleRoutines.C
fearmayo: Done

src/enzo/ActiveParticleRoutines.C
fearmayo: Yes this is the destructor function.

src/enzo/New_Grid_ReadGrid.C
fearmayo: I've had issues with this part of the code alright. I think (from memory) I need to comment these lines out for restarting when no active particles exist. This probably does require an additional PR though. It would be make sense to raise an issue around these lines.

src/enzo/ActiveParticleRoutines.C
fearmayo: Done

src/enzo/ActiveParticleRoutines.C
fearmayo: Done

src/enzo/ActiveParticleRoutines.C
fearmayo: Done

src/enzo/ActiveParticleRoutines.C
fearmayo: I just removed those performance lines.

src/enzo/ActiveParticle_RadiationParticle.C
fearmayo: I'm not so sure. The RadiationParticle is fixed in space in this implementation. Adding another parameter isn't necessarily a good idea. My view is that we leave this as is and if someone wants to use the RadiationParticle and allow it to advect in the future then they can modify this.

src/enzo/ActiveParticle_SmartStar.C
fearmayo: Done

src/enzo/ActiveParticle_SmartStar.C
fearmayo: Done

src/enzo/ActiveParticle_SmartStar.h
fearmayo: I tried that before but it wasn't compatible with the way the AP infrastructure is set up. Having the AccretionRate and AccretionRateTime as arrays works very nicely anyway.

src/enzo/ActiveParticle_SmartStar.h
fearmayo: Typo. Fixed now.

src/enzo/Grid_FindMinimumPotential.C
fearmayo: I believe all the cells in Enzo are cubic so this should be fine.

src/enzo/Grid_FindMassinRegion.C
fearmayo: As above

src/enzo/ActiveParticle_SpringelHernquist.C
fearmayo: Non-intuitive that's for sure. But MassUnits is defined in ActiveParticle.C as 
`data.DensityUnits*POW(data.LengthUnits,3);`

src/enzo/CommunicationShareActiveParticles.C
fearmayo: Indeed this is not good. I think an issue needs to be raised on this.

fearmayo: I've never tested the code with multiple APs. I suspect many parts of the code would break.

src/enzo/CheckOverlap.C
fearmayo: Done

src/enzo/Grid_FindAngularMomentumMinimum.C
fearmayo: Yes you are right! Done

src/enzo/Grid_ApplyGalaxyParticleGravity.C
fearmayo: I think this has to be like this as it's outside the "i" loop

src/enzo/Grid_ActiveParticleHandler.C
fearmayo: Hi Dave, 
Those `flags` values are needed by the active particle type later on. For example the SmartStar particle sets the H2Fraction flag to True. This is then checked/used in `ActiveParticle.C` in the `ConstructData` function. So you can think of the `flags` structure as a booking structure that tells each AP what attributes it has. 
Does that (kinda) make sense?

src/enzo/DetermineSEDParameters.C
fearmayo: Thanks - good catch! Done

src/enzo/Grid_ActiveParticleHandler.C
fearmayo: Hmmm good question. The idea behind this function is that it allows for the conversion of Star Objects into Active Particles. It should be done on the command line similar to the way enzo.exe can create new outputs with the GravitationalPotential for example. I think in this case the usage was ./enzo.exe -a <snapshot> and then the new snapshop would have all of the star objects stored as active particles. It provides backwards compatibility with star objects. However, it looks like the front end part of the code is missing. I may have done that on purpose with the idea of making the frontend a standalone PR. I wrote this bit of code and it worked fine.

src/enzo/ConvertParticles2ActiveParticles.C
fearmayo: I've removed the RebuildHierarchy function and placed DEBUG statements around CollectParticleTypes.

src/enzo/ActiveParticle_SpringelHernquist.C
fearmayo: Fixed

src/enzo/DetermineSEDParameters.C
fearmayo: Fixed

src/enzo/ActiveParticle_SmartStar.C
fearmayo: Done

src/enzo/ActiveParticle_SmartStar.C
fearmayo: I took this fit from other PopIII models in Enzo. I couldn't find a paper that references this fit though. It would be good to include it though.

src/enzo/DetermineSEDParameters.C
fearmayo: I don't think the SLUG PR has made it into Enzo yet.

src/enzo/ActiveParticle_Skeleton.C
fearmayo: The block of code is necessary as an example of how to write an ActiveParticle. I think the `if(false)` should be `if(true)` though so I've changed it. It doesn't matter of course since this is just template code - it never runs.

src/enzo/Grid_AccreteOntoAccretingParticle.C
fearmayo: The accreting particle algorithm has a bug that causes the particle to not conserve momentum. I was able to show that it breaks with the Burkert-Bodenheimer test. I think it's best to leave it for now though as it was the original inspiration for the SmartStar particle and contains a lot of the underlying physics. 

At the moment the only AP I would trust is the SmartStar since it's the only one that has been extensively used/tested. All the other particle types are shells until someone tests/uses them.

src/enzo/New_Grid_WriteGrid.C
fearmayo: Yes :)

src/enzo/ActiveParticle.C
fearmayo: It doesn't appear to be used anyway so I deleted it.

src/enzo/FofLib.C
fearmayo: Yes by the active particle merging function. It is used to find groups for merging.

src/enzo/ActiveParticleFindAll.C
fearmayo: Oh - each ActiveParticleType calls this function separately. So when it returns it returns the number of that type. Is that what you meant?

src/enzo/Grid_ApplyGalaxyParticleGravity.C
ibutsky: Just to clarify - I mean replacing the 0 with "GridStartIndex[0]", not "i".  I suspect that in most cases, GridStartIndex[0] = 0 and this is equivalent anyway

fearmayo: Ah sorry yes. I think you're correct GridStarIndex[0] would be more general

gregbryan: We're talking about the 0 in GRIDINDEX_NOGHOST(o,j,k)?  I think this intended to be a loop over all values, so should start at 0 (just like the j and k loops above).  Note that GridStartIndex is the start of the active region and so generally not 0 (i.e. it is the first non-ghost zone).  So I think this code as originally written is correct...  (but ignore me if I'm talking nonsense!).

src/enzo/hydro_rk/EOS.h
jwise77: Good catch. I confirmed that this should be true. This bug has been in the EOS since the start. I fixed it.

src/enzo/hydro_rk/EOS.h
jwise77: Added an exception

src/enzo/ActiveParticle.h
jwise77: Since it works for now, we won't fix this for the PR. But it should be fixed in the future, so I created issue #74.

src/enzo/ActiveParticle.h
jwise77: Changed.

src/enzo/RebuildHierarchy.C
jwise77: I agree with @fearmayo and it's unclear what's the limit. Nobody's tried to push the limit. It's definitely a good idea to put a warning in the documentation. I'll add this.

jwise77: Actually, there is no dev documentation for the ActiveParticles. Issue #75

src/enzo/ActiveParticleInitialize.C
jwise77: Fixed.

src/enzo/ActiveParticleRoutines.C
jwise77: Clarified.

src/enzo/ActiveParticleRoutines.C
jwise77: This is redundant. I'll remove the "AllMasses" function and make the appropriate changes through the code.

src/enzo/Grid_ApplyGalaxyParticleGravity.C
ibutsky: You're right! I was conflating different use cases. This looks good to me!

src/enzo/New_Grid_ReadGrid.C
jwise77: These lines are written out for analysis (e.g. yt) but aren't needed for a restart, so we skip them.
```C++
fprintf(fptr, "PresentParticleTypes = ");
    if (NumberOfParticles)
      fprintf(fptr, "DarkMatter ");
    for (int i = 0; i<EnabledActiveParticlesCount; i++){
      if (ActiveParticleTypeCount[i] > 0) {
        fprintf(fptr, "%s ", EnabledActiveParticles[i]->particle_name.c_str());
      }
    }
    fprintf(fptr, "\n");
    // And their counts
    fprintf(fptr, "ParticleTypeCounts = ");
    if (NumberOfParticles)
      fprintf(fptr, "%"ISYM" ", NumberOfParticles);
    for (int i = 0; i<EnabledActiveParticlesCount; i++){
      if (ActiveParticleTypeCount[i] > 0) {
        fprintf(fptr, "%"ISYM" ", ActiveParticleTypeCount[i]);
      }
    }
    fprintf(fptr, "\n");
```

src/enzo/CommunicationShareActiveParticles.C
jwise77: Created issue #76

src/enzo/Grid_ActiveParticleHandler.C
jwise77: This function is used in `ConvertParticles2ActiveParticles.C` line 189

src/enzo/ActiveParticle_SmartStar.C
jwise77: Put in a comment

src/enzo/DetermineSEDParameters.C
jwise77: The SLUG PR was declined by the submitter and probably won't be incorporated. Removed the comment.

jwise77:
Okay, I've helped out @fearmayo to address all of the issues from the code review. I'm approving.

PR #59, opened 2019-04-25
Added functionality: new types of moving refinement regions

bwoshea:
This PR expands Enzo's capabilities by adding two main types of functionality and fixes one bug.  The new functionality is:

1.  It expands the functionality of the MustRefineRegion code (which forces a rectilinear portion of a simulation to be refined to at least a user-determined minimum level) so that the user can have a MustRefineRegion whose bounds and minimum level of refinement vary over time/redshift.  This information is specified by a user-supplied data file providing the time/redshift evolution of the MustRefineRegion parameters, and linearly interpolates between entries in this file.

2.  It adds a CoolingRefineRegion option.  This allows the user to specify a rectilinear sub-volume of a simulation in which refinement based on the cooling time can take place, and also vary that over time if desired in the same manner as described above.  The ability to constrain cooling time-based refinement to a subvolume is particularly desirable because in certain environments refinement based on the cooling time can cause an excessive amount of mesh creation, which can cause out-of-memory errors and generally be very computationally expensive.

In addition to these new features, the original moving RefinementRegion code (which just constrains refinement to be in a subvolume, but does not force refinement or restrict it to a specific type of refinement) now linearly interpolates between times in the same manner as the two new moving RefinementRegion.

All of the moving refinement region options now do error checking - if the current simulation time is before or after the times listed in the entries in the user-supplied RefineRegion file, Enzo gives an error message and exits.  This is a highly desirable behavior, because otherwise the refinement region would stay stationary and not give any warnings.


brittonsmith:
@bwoshea, the automated testing stuff has now been merged into the main repo. If you can merge those changes and update the PR, the test suite will run.

doc/manual/source/parameters/index.rst
```diff
@@ -951,10 +951,74 @@ Hierarchy Control Parameters
     refined to at all times. (No default setting)
 ``MustRefineRegionLeftEdge`` (external)
     Bottom-left corner of refinement region. Must be within the overall
-    refinement region. Default: 0.0 0.0 0.0
+    refinement region. If using a moving refinement region, this will
+    ideally correspond to the bottom-left corner in the
```
aemerick: I think the use if "ideally" here is a little misleading... Maybe "...this should correspond..." or "...this should be set to the bottom-left..."....

doc/manual/source/parameters/index.rst
```diff
@@ -951,10 +951,74 @@ Hierarchy Control Parameters
...
+    CoolingRefineRegion is a rectilinear region in which refinement
+    can be based on the cooling time  (``CellFlaggingMethod`` 7 must be
+    set) but refinement based on the cooling time will not occur
+    outside of this region. Default: 0
+``EvolveCoolingRefineRegion`` (external)
+    1 if the CoolingRefineRegion is evolving; 0 if not. Default: 0
+``CoolingRefineRegionLeftEdge`` (external)
+    Bottom-left corner of refinement region. Must be within the overall
+    refinement region.  If using a moving refinement region, this will
+    ideally correspond to the bottom-left corner in the
```
cbrummelsmith: As Andrew said above, the word "ideally" is a bit confusing. If there are situations where the LeftEdge does not correspond to the bottom-left corner, that should be made clear. If not, then the word 'ideally' is not needed.

src/enzo/FindSubgrids.C
```diff
@@ -67,18 +67,18 @@ int FindSubgrids(HierarchyEntry *Grid, int level, int &TotalFlaggedCells,
...
- 
+
+  /* Flag cells based on MultiRefineRegions */
+
+  if (CurrentGrid->SetFlaggingFieldMultiRefineRegions(level)
+      == FAIL) {
+    fprintf(stderr, "Error in grid->SetFlaggingFieldMultiRefineRegions.\n");
+    return FAIL;
+  }
+
```
aemerick: This is probably OK, but why did this get moved outside of the `NumberOfFlaggedCells !=0` if statement?

aemerick:
PR looks good overall. Just had a couple minor comments / questions.

src/enzo/SetEvolveRefineRegion.C
```diff
@@ -23,68 +27,281 @@ int SetEvolveRefineRegion (FLOAT time)
...
+	  break;
+	}
+      }
+    }else{  // code time
+
+      /* Check to see if the current time is within the bounds of the given time file.  This assumes
+	 that EvolveRefineRegionTime[0] is the earliest time and the last output is the latest time (for
+	 the given track file). */
+      if(time < EvolveRefineRegionTime[0] || time > EvolveRefineRegionTime[EvolveRefineRegionNtimes-1]){
+	fprintf(stderr,"SetEvolveRefineRegion ERROR: current simulation time is outside of range of track file redshifts!");
```
cbrummelsmith: This print statement should say "times" instead of "redshifts".
```suggestion
	fprintf(stderr,"SetEvolveRefineRegion ERROR: current simulation time is outside of range of track file times!");
```

src/enzo/SetEvolveRefineRegion.C
```diff
@@ -23,68 +27,281 @@ int SetEvolveRefineRegion (FLOAT time)
...
-      if( time > EvolveRefineRegionTime[timestep] ){
-        break;
+  /* does time evolution for standard refinement regions */
+  if((RefineRegionTimeType == 1) || (RefineRegionTimeType == 0) ){
+  
+    /* Find closest time step with <time */
+    /* Set time=redshift if that's what we're doing. */
+    if (RefineRegionTimeType == 1) {  // redshift
+
+      time = redshift;
```
cbrummelsmith: This could be problematic if ComovingCoordinates is False and RefineRegionTimeType == 1. In that case, redshift would not be defined. The user shouldn't be using RefineRegionTimeType == 1 in a non-cosmological simulation but it is possible.

aemerick: Good catch. Maybe just adding an error check in `ReadParameterFile`?

src/enzo/SetEvolveRefineRegion.C
```diff
@@ -23,68 +27,281 @@ int SetEvolveRefineRegion (FLOAT time)
...
+	  break;
+	}
+      }
+    }else{ // code time
+
+      /* Check to see if the current time is within the bounds of the given time file.  This assumes
+	 that EvolveMustRefineRegionTime[0] is the earliest time and the last output is the latest time (for
+	 the given track file). */
+      if(time < EvolveMustRefineRegionTime[0] || time > EvolveMustRefineRegionTime[EvolveMustRefineRegionNtimes-1]){
+	fprintf(stderr,"SetEvolveRefineRegion ERROR for EvolveMustRefineRegion: current simulation time is outside of range of track file redshifts!");
```
cbrummelsmith:
```suggestion
	fprintf(stderr,"SetEvolveRefineRegion ERROR for EvolveMustRefineRegion: current simulation time is outside of range of track file times!");
```

src/enzo/SetEvolveRefineRegion.C
```diff
@@ -23,68 +27,281 @@ int SetEvolveRefineRegion (FLOAT time)
...
+      }
     }
+    timestep -= 1;
+    if (timestep < 0) return SUCCESS;
+
+    /* Set MustRefineRegion to EvolveMustRefineRegion */
+
+    for (i = 0; i < MAX_DIMENSION; i++){
+
+      if(timestep==EvolveMustRefineRegionNtimes-1){
```
cbrummelsmith: Small spacing change
```suggestion
      if(timestep == EvolveMustRefineRegionNtimes-1){
```

src/enzo/SetEvolveRefineRegion.C
```diff
@@ -23,68 +27,281 @@ int SetEvolveRefineRegion (FLOAT time)
...
-  for (int flag_method = 0; flag_method < MAX_FLAGGING_METHODS; flag_method++) {
-    if( CellFlaggingMethod[flag_method] == 12 ){
-      for (int dim = 0; dim < MAX_DIMENSION; dim++){
-        MustRefineRegionLeftEdge[dim]=EvolveRefineRegionLeftEdge[timestep][dim];
-        MustRefineRegionRightEdge[dim]=EvolveRefineRegionRightEdge[timestep][dim];
+      /* Check to see if the current time is within the bounds of the given time file.  This assumes
+	 that EvolveRefineRegionTime[0] is the earliest time and the last output is the latest time (for
+	 the given track file). */
+      if(time < EvolveCoolingRefineRegionTime[0] || time > EvolveCoolingRefineRegionTime[EvolveCoolingRefineRegionNtimes-1]){
+	fprintf(stderr,"SetEvolveRefineRegion ERROR: current simulation redshift is outside of range of track file redshifts! (Cooling time box)");
```
cbrummelsmith:
```suggestion
	fprintf(stderr,"SetEvolveRefineRegion ERROR: current simulation redshift is outside of range of track file times! (Cooling time box)");
```

doc/manual/source/parameters/index.rst
bwoshea: Thanks, I'll fix that!

doc/manual/source/parameters/index.rst
bwoshea: Yep, I will fix that.

src/enzo/FindSubgrids.C
bwoshea: This is done because there's a conflict between MustRefineParticles and MultiRefineRegions.  The MustRefineParticles are set up to derefine everywhere outside of their area, so if you want to force refinement in that area you have to do it separately.  That's why this is here.

src/enzo/SetEvolveRefineRegion.C
cbrummelsmith:
```suggestion
	fprintf(stderr,"SetEvolveRefineRegion ERROR: current simulation time is outside of range of track file times! (Cooling time box)");
```

src/enzo/SetEvolveRefineRegion.C
bwoshea: Fixed!

src/enzo/SetEvolveRefineRegion.C
bwoshea: Done!

src/enzo/FindSubgrids.C
kohdaegene: reading through, it doesn't look like it should've been in there in the first place?

PR #69, opened 2019-06-11
CircleCI tag change, MHD Orszag-Tang bugfix

bwoshea:
This makes two small changes:

1.  The CircleCI config.yml file has been updated so that it fetches tags from the main repository.  This is to ensure that any fork of Enzo, no matter how long ago it was forked off, will get the tags corresponding to the Gold Standard.  (Tags are not fetched by default when you do a 'git pull' from the main repository.)

2.  The MHD Orszag-Tang test problem has a refinement track file that ended before the simulation ended, which caused the code to deliberately crash and issue an error message.  I've updated the test problem so that there's an additional line in the parameter file, which will return it to the expected behavior of not crashing.


.circleci/config.yml
```diff
@@ -9,8 +9,8 @@ commands:
           echo 'export ENZOTEST_DIR=$HOME/enzo_test' >> $BASH_ENV
           echo 'export ANSWER_NAME="push_suite"' >> $BASH_ENV
           echo 'export HYPRE=hypre-2.11.2' >> $BASH_ENV
-          # tag the tip so we can go back to it
-          git fetch --tags origin
+          # get tags from the main repository (for the current gold standard)
+          git fetch --tags https://github.com/enzo-project/enzo-dev
           git tag tip
```
brittonsmith: We should move the original comment down to the line below.
```suggestion
          # tag the tip so we can go back to it
          git tag tip
```

PR #71, opened 2019-06-11
Domain flux

aemerick:
This enables a counter to keep track of the mass that outflows from the simulation boundary for both the total density and the species fields (chemistry and metallicity). This is stored cumulatively directly in the parameter files, and also cycle-by-cycle in a text file.


src/enzo/Grid_ComputeDomainBoundaryMassFlux.C
```diff
@@ -0,0 +1,140 @@
...
+
+  /* Return if this doesn't concern us */
+
+  if (ProcessorNumber != MyProcessorNumber)
+    return SUCCESS;
+
+  PrepareBoundaryMassFluxFieldNumbers(); //  this should be put somewhere else
+
+  /* get units */
+  const float msolar = 1.989e33;
```
jwise77: Use value from `phys_constants.h`

src/enzo/Grid_ComputeDomainBoundaryMassFlux.C
```diff
@@ -0,0 +1,140 @@
...
+  if (MultiSpecies > 0)
+    NumberOfBoundaryMassFields += 5; // HI, HII, HeI, HeII, HeIII
+
+  if (MultiSpecies > 1)
+    NumberOfBoundaryMassFields += 3; // H2, H2I, HM
+
+  if (MultiSpecies > 2)
+    NumberOfBoundaryMassFields += 3; // D, D+, HD
+
+  if (MultiMetals > 0)
```
jwise77: `MultiMetals` isn't always set to be >0 when metal fields are being used and multiple metal fields might be used. You could use `grid::IdentifyColourFields` to count the metal fields when the simulation starts and then store this number somewhere.

src/enzo/Grid_PrepareBoundaryMassFluxFieldNumbers.C
```diff
@@ -0,0 +1,91 @@
...
+  int DeNum, HINum, HIINum, HeINum, HeIINum, HeIIINum, HMNum, H2INum, H2IINum,
+      DINum, DIINum, HDINum;
+  if ( MultiSpecies ){
+    IdentifySpeciesFields(DeNum, HINum, HIINum, HeINum, HeIINum, HeIIINum,
+                          HMNum, H2INum, H2IINum, DINum, DIINum, HDINum);
+  }
+
+  /* get metallicity tracer field number */
+  int MetalNum;
+  MetalNum   = FindField(Metallicity, this->FieldType, this->NumberOfBaryonFields);
```
jwise77: You could use `grid::IdentifyColourFields` here.

aemerick:
I've updated this PR with the suggested improvements. Should be good to go now.

bwoshea:
@aemerick , I still see two issues:

1) In Grid_ComputeDomainBoundaryMassFlux.C , you need to define MassUnits as double instead of float (which is how it's done in the rest of the code).  If one compiles with 32-bit precision or mixed 32/64-bit precision, having a 32-bit MassUnits may cause arithmetic overflow, depending on the size of the box.

2) You have a conflict in Grid.h that needs to be resolved.

If you can fix those two, I think we can go ahead and approve this!

bwoshea:
Looks good to me!  I'm approving and merging.

PR #78, opened 2019-06-12
README update

jwise77:
Removing G+ reference and one last mercurial mention in README

### DIFF
--- a/README
+++ b/README
@@ -16,7 +16,7 @@ Enzo is developed in the open on github.com:
 
 Documentation, including instructions for compilation, can be found at:
 
- * http://enzo-project.org/docs/2.5/
+ * http://enzo-project.org/docs/2.6/
 
 Please subscribe to the Enzo Users' mailing list at:
 
--- a/README
+++ b/README
@@ -27,12 +27,8 @@ the Enzo Developer's mailing list as well:
 
  * https://groups.google.com/forum/#!forum/enzo-dev
 
-You can also follow the Enzo Project on google+ at:
-
- * https://plus.google.com/115923030596894217717
-
 If you have received this source code through an archive, rather than the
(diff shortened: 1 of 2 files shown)


jwise77:
> Looks good to me. This could also be a good opportunity to put in all of the new contributors and update email addresses in the README, but that's certainly not necessary.

I think that's a good idea to consolidate changes into this PR. I'll search the commit log for people not in the list.

jwise77:
Okay, I went through the commit log and added the missing authors.

bwoshea:
There are three approves, but I propose that we hold off until we're ready to release Version 2.6 - that way we can update the docs and also update the link to the docs (line 19 in the README file).

README
```diff
@@ -45,33 +41,43 @@ list of the people who have recently contributed, in alphabetical order:
...
-   * Shuo Kong		    skong.astro@gmail.com
+   * Shuo Kong              skong.astro@gmail.com
+   * Claire Kopenhafer      clairekope@gmail.com
    * Alexei Kritsuk         akritsuk@ucsd.edu
    * Michael Kuhlen         kuhlen@gmail.com
    * James Larrue           james.larrue@diopolis.com
    * Eve Lee                elee@cita.utoronto.ca
-   * Miao Li		    ml3322@columbia
+   * Miao Li                ml3322@columbia.edu
    * Yuan Li                yuan@astro.columbia.edu
```
gregbryan: Assuming the FDM module is accepted, can we add it's author:

```suggestion
    * Yuan Li                yuan@astro.columbia.edu
    * Xinyu Li               xinyu.li@columbia.edu
```

jwise77: Done in a separate commit

bwoshea:
OK, we're good to go here - we're going to send the email out slightly later today!

Issue #79, opened 2019-06-17
Problems restarting cosmology simulations following Active Particles pull request


bwoshea:
I'm having problems restarting an old Enzo cosmological simulation following the acceptance of the active particle PR (PR #57 ).  When I restart a small test simulation (found at https://www.dropbox.com/s/hbhmhijfcgb8y8e/RD0006.tar.gz?dl=0) with the command line:

> ./enzo.exe -d -r RD0006/RD0006

I get this error:

> OPEN data hierarchy RD0006/RD0006.hierarchy
> Caught fatal exception:
> 
>    'Error reading ParticleFileName.'
> at New_Grid_ReadGrid.C:216
> 
> Backtrace:
> 
> BT symbol: 0   enzo.exe                            0x000000010c0bf4a2 _ZN18EnzoFatalExceptionC2EPKcS1_i + 66
> BT symbol: 1   enzo.exe                            0x000000010c2947b0 _ZN4grid14Group_ReadGridEP7__sFILEixPciibi + 7760
> BT symbol: 2   enzo.exe                            0x000000010c324d0d _Z23Group_ReadDataHierarchyP7__sFILEiP14HierarchyEntryR11TopGridDataiS2_iiPibS0_ + 637
> BT symbol: 3   enzo.exe                            0x000000010c324328 _Z17Group_ReadAllDataPcP14HierarchyEntryR11TopGridDataP16ExternalBoundaryPdb + 1624
> BT symbol: 4   enzo.exe                            0x000000010c0be1a4 main + 948
> BT symbol: 5   libdyld.dylib                       0x00007fff7b901015 start + 1
> BT symbol: 6   ???                                 0x0000000000000004 0x0 + 4

I am confident that it's this PR because when I rewind the repository a few days (running from when we merged PR #63 , changeset 7c0965138124d55f20463f1cc0ae6951cd58d837) things work just fine. 

As I dig into this, it seems that the problem is lines 208-209 in file src/enzo/New_Grid_ReadGrid.C:

>     // We read these in but don't use them (yet?). Therefore we won't error
>     // out if we can't read them in.
>     fgets(unused_string, MAX_LINE_LENGTH, fptr);
>     fgets(unused_string, MAX_LINE_LENGTH, fptr);

If I comment out the two fgets lines, it works fine, but that will cause problems for simulations with active particles.  The better fix seems to be to add this logic right around those lines:

>     int ActiveParticlesExist = 0;   <------ NEW CODE
>     
[...]

PR #80, opened 2019-06-17
Small tweak to hierarchy file reader - patch for pre-AP simulations

bwoshea:
This is described by the discussion in Issue #79 , and implements the fix described there.  It fixes a bug where the recently-accepted active particle PR (PR #57 ) breaks restarts from simulations that have particles but predate the active particle PR merge (which is to say, almost all Enzo simulations ever).

### DIFF
--- a/src/enzo/New_Grid_ReadGrid.C
+++ b/src/enzo/New_Grid_ReadGrid.C
@@ -198,27 +198,18 @@ int grid::Group_ReadGrid(FILE *fptr, int GridID, HDF5_hid_t file_id,
             ENZO_FAIL("error reading NumberOfParticles.");
     }
 
-    int ActiveParticlesExist = 0;
-
-     if (fscanf(fptr, "NumberOfActiveParticles = %"ISYM"\n", &NumberOfActiveParticles) == 1) {
-       fgets(unused_string, MAX_LINE_LENGTH, fptr);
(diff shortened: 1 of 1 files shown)


src/enzo/New_Grid_ReadGrid.C
```diff
@@ -198,15 +198,23 @@ int grid::Group_ReadGrid(FILE *fptr, int GridID, HDF5_hid_t file_id,
             ENZO_FAIL("error reading NumberOfParticles.");
     }
 
+    int ActiveParticlesExist = 0;
+
     if (fscanf(fptr, "NumberOfActiveParticles = %"ISYM"\n", &NumberOfActiveParticles) != 1) {
```
brittonsmith: The logic here is a bit confusing. Could it not be done like this?
```suggestion
    if (fscanf(fptr, "NumberOfActiveParticles = %"ISYM"\n", &NumberOfActiveParticles) == 1) {
      fgets(unused_string, MAX_LINE_LENGTH, fptr);
      fgets(unused_string, MAX_LINE_LENGTH, fptr);
    } else {
      NumberOfActiveParticles = 0;
    }
```

bwoshea: This seems like a much cleaner implementation to me.  Thanks!

bwoshea:
Followup: I just pushed some changes to clean up the code, following @brittonsmith 's suggestion.

PR #84, opened 2019-06-19
Fixing MinimumPressureSupport in hydro_rk solvers

ibutsky:
The implementation of the MinimumPressureSupportParameter in HydroMethod = 3 and 4 was inconsistent between the two HydroMethods, had hard-coded parameters, and didn't match the implementation in the Zeus solver. When I ran a cosmological simulation with MinimumPressureSupport and the hydro solvers, I got some pretty unphysical values in the temperature and gas energy.  This fix keeps the original framework of the MinimumPressureSupport implementation found in the hydro_rk/ files, but calculates the "min_coeff" parameter with Grid_SetMinimumSupport for consistency.


bwoshea:
This seems like a fairly straightforward update.  Nice job!  I'm glad you made this all consistent.

dcollins4096:
That looks good to me.  Thanks for tracking this down!

PR #87, opened 2019-08-02
updating changelog for v2.6

kohdaegene:


bwoshea:
I'm merging without three approves, since this is a simple update and we also already sent out the announcement email...

PR #88, opened 2019-08-05
Switch configure script to bash.

brittonsmith:
I'm open to this being rejected outright, but I thought since our make system uses bash we might as well be consistent with that. Also, fewer people use csh anymore and so you frequently have to install it.


jwise77:
I'm +1 on this change, especially since csh isn't installed by default on many distributions.

PR #89, opened 2019-08-05
Add test of various compile options.

brittonsmith:
I added a couple nested loops to test different combos of baryon/particle precision and integers/particle id precision. So to not do too many, I skip the combinations where the precision is the same. There is also a check to see if Enzo compiles without mpi.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -145,7 +145,7 @@ commands:
           make -j 4 || (make show-config ; exit 1)
           make use-mpi-yes
           # test ray-tracing
-          make use-photon-no
+          make photon-no
           make clean
           make -j 4 || (make show-config ; exit 1)
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -105,6 +105,7 @@ commands:
       - run: |
           source $BASH_ENV
           source $HOME/venv/bin/activate
+          alias build-enzo='make -j 4 || (make show-config ; exit 1)'
           ./configure
           cd src/enzo
           make machine-linux-gnu
@@ -119,7 +120,7 @@ commands:
           make precision-$prec
           make particles-$part
           make clean
-          make -j 4 || make show-config
+          build-enzo
           done ; done
           # return this to default
           make precision-64
@@ -135,12 +136,12 @@ commands:
           make integers-$inte
           make particle-id-$pids
           make clean
-          make -j 4 || make show-config
+          build-enzo
           done ; done
           # test without mpi
           make use-mpi-no
           make clean
-          make -j 4 || make show-config
+          build-enzo
 
   build-docs:
     description: "Test the docs build."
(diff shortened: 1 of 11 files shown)


.circleci/config.yml
```diff
@@ -99,6 +99,47 @@ commands:
...
+          source $HOME/venv/bin/activate
+          ./configure
+          cd src/enzo
+          make machine-linux-gnu
+          make load-config-test-suite
+          # test various baryon/particle precisions
+          for prec in 32 64
+          do for part in 32 64 128
+          do
+          if [ $prec == $part ]; then
```
jwise77: I think we should still test the precision-32 and particles-32 case. But we shouldn't test the 64/64 case because that's the default.

.circleci/config.yml
```diff
@@ -99,6 +99,47 @@ commands:
...
+          make clean
+          make -j 4
+          done ; done
+          # return this to default
+          make particles-64
+          # test various integer/particle-id precisions
+          for inte in 32 64
+          do for pids in 32 64
+          do
+          if [ $inte == $pids ]; then
```
jwise77: Same comment as above

brittonsmith: It looks like the test-suite config is the following:
```
   CONFIG_PRECISION  [precision-{32,64}]                     : 64
   CONFIG_PARTICLES  [particles-{32,64,128}]                 : 64
   CONFIG_INTEGERS  [integers-{32,64}]                       : 32
   CONFIG_PARTICLE_IDS  [particle-id-{32,64}]                : 32
```
I'll change it to skip that one.

.circleci/config.yml
```diff
@@ -99,6 +99,47 @@ commands:
...
+          if [ $prec == $part ]; then
+          continue
+          fi
+          make precision-$prec
+          make particles-$part
+          make clean
+          make -j 4
+          done ; done
+          # return this to default
+          make particles-64
```
gregbryan: I would add "make precision-64" to return that to default as well (I know the above loop should leave it there, but what if that loop got changed) -- also this just makes it clearer.

gregbryan:
I'm definitely for this as long as it doesn't take too long to test all of those combinations.

.circleci/config.yml
brittonsmith: Sure thing.

brittonsmith:
The test-suite takes about 45 minutes to run and I think it's about 5 minutes per build of enzo. The compile test and the test-suite will run simultaneously, so I don't think we should be adding too much time overall.

Note, that's me manually killing the test-suite since we know the compile tests are failing right now.

brittonsmith:
@jwise77, @gregbryan, adding this caught a number of compile issues with various settings. It's probably worth another go through.

jwise77:
> @jwise77, @gregbryan, adding this caught a number of compile issues with various settings. It's probably worth another go through.

I went through the latest changes, and LGTM!

gregbryan:
Three approvals, tests all pass -- I'm merging.

PR #90, opened 2019-08-05
BUGFIX: precisions changes for non-default compilation

gregbryan:
These are some changes to recently committed code to get it to compile for non-default precision compilation (precision-32).  Thanks to @fearmayo for pointing these out.


src/enzo/Grid_SchrodingerSolver.C
```diff
@@ -63,22 +63,17 @@ int rk3tvd(float *repsi, float *impsi, float *repsi0, float *impsi0,
 	       float hmcoef,
          int gravity, double *p);
 
-int rk4(double *repsi, double *impsi,
+int rk4(float *repsi, float *impsi,
          int in, int jn, int kn, int rank,
-         double dt, double dx[], double dy[], double dz[],
+         float dt, float dx[], float dy[], float dz[],
```
fearmayo: Since dx is a position (well distance) should it be FLOAT also? I think this is the cell width in this case.

jwise77: Distance differences are usually `float` because it's only relative differences, not absolute. I'm not sure why `CellWidth` is a `FLOAT`. I remember it being a normal `float` at some point in the past.

fearmayo: Ah ok. Then it's fine. `CellWidth` is a float here so that's probably fine.

gregbryan: I also forget why `CellWidth` was promoted.  I think the use of `float` here is OK, but happy to change if anyone feels strongly.

bwoshea: No strong feeling on this.  In C++ floats get promoted to doubles if anything on the right hand side of the expression is a double.  So, it seems fine in general!

Issue #101, opened 2019-08-11
CUDA docs seem a bit out of date


zingale:
Not part of the JOSS review, but I managed to build Enzo with CUDA and run on a Titan V card.  The docs suggest using `sm_20`, but that isn't even recognized by later versions of CUDA.  I ran with CUDA 10.1 and `sm_70` on the Titan V and it seemed to work (e.g. `nvidia-smi` showed the `./enzo` executable running on the GPU)

Perhaps the docs should be updated with suggestions on more recent cards?

PR #111, opened 2019-09-11
[JOSS] Update Cuda documentation

pgrete:
Addresses #101 

Also removed `CONFIG_OBJECT_MODE` along the line as it potentially confuses the user as the last "64" being left in `make show-config` after setting all settings to single precision as suggested by the CUDA related doc.


gregbryan:
This looks good, but I think there are still some (harmless) references to object-mode (e.g. line 80/81 and 265 of Make.config.target, and line 22 and 118-122 of Make.config.settings).

pgrete:
Good catch. Just removed those references.

Issue #96, opened 2019-08-11
[JOSS REVIEW] linux-gnu build info is distro specific


zingale:
Upon running 
```
make machine-linux-gnu
```
it outputs:
```
 *** Execute 'gmake clean' before rebuilding executables ***

   MACHINE: Use apt-get to install libhdf5-serial-dev gfortran openmpi-bin libopenmpi-dev
   MACHINE-NAME: linux-gnu
```

but many Linux distros do not use `apt-get` and new users could be confused.  It would be helpful to have a list of the requirements needed for Enzo in the README.md or a REQUIREMENTS.txt

zingale:
JOSS Review: https://github.com/openjournals/joss-reviews/issues/1636

bwoshea:
@zingale  This is addressed in PR #115

zingale:
looks good!

PR #115, opened 2019-09-11
[JOSS] Requirements

pgrete:
Addresses #96 

Adds info on requirements already at the README level and make the instructions for GNU/Linux more generic.

I also renamed `README` to `README.md` and updated the content to use markdown so that it's properly rendered on GitHub.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,7 +46,7 @@ commands:
           cd $HOME/grackle
           ./configure
           cd src/clib
-          make machine-ubuntu
+          make machine-linux-gnu
           make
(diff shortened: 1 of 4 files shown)


brittonsmith:
I think having a Debian/Ubuntu specific make file is fine since it is specifically for building with packages installed by apt-get. As well, the package locations and names have changed a bit over time, so it's probably better to keep that in a separate make file rather than to keep updating the standard linux example.

bwoshea:
I concur as well, @pgrete ; your proposed solution makes sense to me too.

PR #126, opened 2019-10-25
WIP - Individual stars

aemerick:
------------- Work in Progress ---------------

A large PR containing the individual star formation, stellar feedback, and chemical evolution routines (described in more detail here: [Emerick+2019](https://ui.adsabs.harvard.edu/abs/2019MNRAS.482.1304E/abstract)). These routines piggy back on the already-existing StarParticle class.

This is marked as a work in progress because the code needs a bit of cleaning. Namely old routines no longer in use, bits of code where I prioritized time-to-implementation over generalizeability / style, and other bits of code that probably shouldn't get pulled into the main repo. In addition, I need to add documentation for all of this. 

Creating this as a way to more cleanly look at all of the changes and assess what needs work. 

New features and to-do's (incomplete list... adding as I go.....):

0) **To Do**: Later in development I started adding compile-time ifdef for individual stars due to some more invasive changes I had to make for a variety of reasons. Need to decide to use this ifdef or not, and if so, retroactively add the rest of the code to this.

1) Star formation and stellar feedback with stars representing individual star particles

2) Feedback includes: AGB winds, massive star stellar winds, core collapse SNe, Type Ia SNe, ionizing radiation coupled to ray tracing, and Lyman-Werner and FUV radiation (for photoelectric heating) with options to use pure optically-thin mode (1/r^2 profile) OR ray tracing. Feedback is deposited in spherical region (thermal only). StarParticle class use allows this to work across grid boundaries. **To Do:** Star properties are determined with use of ascii input tables that can (and probably should) be converted to a single non-ascii file.

3) Stars are coupled to follow multi-element yields (chosen arbitrarily at run-time) given an input yield table. Stars are chemically tagged at birth and abundances written to file (optional) or tracked as an attribute (optional, memory intensive). **To Do**: Ensure table-reader is general to allow for use of different yield tables at run-time.

4) Includes PopIII star model. Near copy of existing model, but with some edits to make it self-consistent. Includes multi-element PopIII yields.

5) Option to follow "unresolved stars" and set a particle mass floor below which stars still form but are lumped into a single particle to cut down on number of particles. 

6)  Extra metal fields (optional) used to track origin of elements: AGB, core collapse SNe, Type Ia SNe, and PopIII (some of this already existed).

7) Many additions to GalaxySimulation ICs (**To Do**: Document these). Cosmology and Agora problem types also have additions to work with yield model **Possible To Do**: Generalize metal field initialization to look more like how the RT fields are initialized so this doesn't have to get copied across many initializers.

8) Generalized static/external background gravity model using a double power law profile permitting several specific cases (e.g. plummer, sphere, NFW, Moore, Hernquist, Jaffe). **To Do**: Much documentation.

9) **To Do**: Clean up the (many) run time parameters in this PR to what is actually needed. Adjust names on a few for clarity.


jwise77:
We updated the gold standard after #131, which also created some conflicts with this PR. Could you merge the latest changes from enzo-project:master so the tests can run again? Thanks!

aemerick:
I'm closing out this PR in favor of a new one (coming very soon). Made a lot of changes since I last touched this and it'll be easier to just issue a new one.

PR #127, opened 2019-11-08
Optically Thin Source Clustering

aemerick:
Removing an ifdef in favor of a run-time parameter to turn on / off source clustering with optically thin radiation. 

A few years ago I had tested this routine and found it to give strange results, but I recently tried to use this again and it looks like it is behaving well. I'm not sure what has changed between then and now that fixed this, but none of the commit logs in the clustering routines seem to be more recent than when I first looked at this. Its possible there was some issue in an unrelated piece of code from my individual stars methods that affected this that has since been fixed, but tracking this down would be hard. 

I added a word of caution to the docs on using this piece of code just in case weird things happen again. It speeds up computation considerably (> factor of 10) when using optically thin radiation (1/r^2 profiles) with many sources / grid cells.

My test simulation was a restart of one of my isolated dwarf galaxy runs with individual stars. At this particular point there were ~160 radiation sources. I'm attaching slices comparing the H2I disassociation rate computed using the default cell-by-cell / star-by-star 1/r^2 deposition and the source clustering algorithm, along with a radial profile comparison between the two. There are clearly errors very close to individual sources, but these are not large (**Edit:** I was somewhat wrong about how large they are, see added plots in comments).

Cell-by-cell (more expensive / accurate).... black points are sources
cell-by-cell

Optically thin source clustering with `RadiativeTransferPhotonMergeRadius=3`
source-clustering

radial_profile

I did find that the behavior is very sensitive to the value of `RadiativeTransferPhotonMergeRadius`. Smaller values (<5) seem to do well with not much error, but larger values (e.g. 10) smooth out the radiation near sources considerably. I added a note of caution about this.

I also changed the docs to reflect that the default value of `RadiativeTransferPhotonMergeRadius` is actually 10. And because I made an oopsie, there are two very minor commits in this PR that are also in PR#120

### DIFF
--- a/doc/manual/source/parameters/index.rst
+++ b/doc/manual/source/parameters/index.rst
@@ -2729,7 +2729,7 @@ Radiative Transfer (Ray Tracing) Parameters
     which is the luminosity weighted center of two sources. This radius
     is in units of the separation of two sources associated with one
     SuperSource. If set too small, there will be angular artifacts in
-    the radiation field. Default: 10.0
+    the radiation field. Default: 3.0
 ``RadiativeTransferSourceBeamAngle`` (external)
     Rays will be emitted within this angle in degrees of the poles from sources with "Beamed" types.  Default: 30
 ``RadiativeTransferPeriodicBoundary`` (external)
--- a/doc/manual/source/parameters/index.rst
+++ b/doc/manual/source/parameters/index.rst
@@ -2714,12 +2714,22 @@ Radiative Transfer (Ray Tracing) Parameters
 ``RadiativeTransferSourceClustering`` (external)
     Set to 1 to turn on ray merging from combined virtual sources on a
     binary tree. Default: 0.
+``RadiativeTransferOpticallyThinSourceClustering`` (external)
+    Set to 1 to use the above tree when computing optically thin radiation
+    (``RadiativeTransferSourceClustering`` must be set to 1), which can greatly
+    reduce computational time for many sources / large cell counts. If on, this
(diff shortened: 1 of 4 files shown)


aemerick:
Better quantification of errors around individual sources. Looks like they are quite large (up to a factor of 5-8 or so) in the host cells of individual sources, but these are only a few cells. This is potentially an issue in some applications, but since this is likely to be used in conjunction with RT of HI and HeI, heating from that (and therefore H2 dissociation) will likely be dominant in those cells anyway:

Star-by-star / cell-by-cell
star_by_star_profile

Source Clustering
source_clustering_profile

fearmayo:
This looks fine but is it worth updating the `RadiativeTransferPhotonMergeRadius` to something like 3?

jwise77:
HI @aemerick, I agree with John R's suggestion about the default merge radius. @fearmayo after this change is made, can you approve? Thanks!

gregbryan:
This has three approvals, passes tests and I think all requested changes were made -- any reason not to merge?

aemerick:
@gregbryan Its ready as far as I'm concerned.

I see that I have the ability to merge the PR with the main branch. Is it fine for the issuer of the PR to do this or better to have one of the reviewers / PR-czar do the merging?

gregbryan:
> @gregbryan Its ready as far as I'm concerned.
> 
> I see that I have the ability to merge the PR with the main branch. Is it fine for the issuer of the PR to do this or better to have one of the reviewers / PR-czar do the merging?

Better to have someone else do it.  I'll do it right now.

PR #130, opened 2019-12-09
updated Make.mach.sdsc-comet to new compilers/hdf5 version

nsimakov:
SDSC have updated their software stack and removed old compilers.

So I updated src/enzo/Make.mach.sdsc-comet to use intel/2018.1.163 mvapich2_ib/2.3.2 hdf5/1.8.21.

Enzo, inits and ring was successfully compiled and executed on one example.


nsimakov:
I have merged with recent updates and fixes. Test have passed. I also compiled enzo on comet and one run example.

src/enzo/Make.mach.sdsc-comet
```diff
@@ -92,8 +96,8 @@ MACH_OPT_AGGRESSIVE  = -O3 -march=native # Flags for aggressive optimization
 # Includes
 #-----------------------------------------------------------------------
 
-LOCAL_INCLUDES_MPI    = 
-LOCAL_INCLUDES_HDF5   = -I$(LOCAL_HDF5_INSTALL)/include # HDF5 includes
+LOCAL_INCLUDES_MPI    =  
+LOCAL_INCLUDES_HDF5   = -I$(LOCAL_MPI_INSTALL)/include -I$(LOCAL_HDF5_INSTALL)/include # HDF5 includes
```
brittonsmith: Should the MPI includes be moved to the line above?

nsimakov: LOCAL_INCLUDES_HDF5 should have -I$(LOCAL_MPI_INSTALL)/include because parallel hdf5 library on comet and this way tools like inits would not compile. So the question is should it be duplicated in LOCAL_INCLUDES_HDF5

brittonsmith: I wasn't aware that Enzo worked with parallel-enabled hdf5. If that's so, then I think this is just fine. Is there a serial version of hdf5 on Comet?

nsimakov: Yes, they do. I tried hdf5-serial/1.10.3 and it works fine. Update Make.mach.sdsc-comet  on that.

PR #131, opened 2019-12-10
Bugfix running with Grackle compiled but Grackle off

aemerick:
Fixes bugfix in mapping Grackle's default parameters to the equivalent Enzo parameters when Grackle is compiled but not in use. This now only occurs when Grackle is actually in use.

I threw this together quickly and just tested that this fixes the crashes that happened in  PhotonTestAMR. So worth a look.

(This PR includes a couple very minor (but fully unrelated) commits from another currently open but accepted PR)

### DIFF
--- a/src/enzo/ReadParameterFile.C
+++ b/src/enzo/ReadParameterFile.C
@@ -48,6 +48,7 @@ int ReadListOfFloats(FILE *fptr, int N, float floats[]);
 int ReadListOfInts(FILE *fptr, int N, int nums[]);
 int CosmologyReadParameters(FILE *fptr, FLOAT *StopTime, FLOAT *InitTime);
 int GrackleReadParameters(FILE *fptr, FLOAT InitTime);
+int GrackleSetDefaultParameters(FILE *fptr);
 int ReadUnits(FILE *fptr);
 int InitializeCosmicRayData();
 int InitializeRateData(FLOAT Time);
@@ -79,6 +80,15 @@ int ReadParameterFile(FILE *fptr, TopGridData &MetaData, float *Initialdt)
   char **active_particle_types;
   active_particle_types = new char*[MAX_ACTIVE_PARTICLE_TYPES];
   int active_particles = 0;
+
+#ifdef USE_GRACKLE
+  /* Check if use_grackle is True and copy over default parameters to
+     their Enzo equivalents */
+  if (GrackleSetDefaultParameters(fptr) == FAIL){
+    ENZO_FAIL("Error in GrackleSetDefaultParameters.\n");
+  }
+  rewind(fptr);
+#endif // USE_GRACKLE
 
   /* read until out of lines */
 
@@ -1622,7 +1632,7 @@ int ReadParameterFile(FILE *fptr, TopGridData &MetaData, float *Initialdt)
   // make sure that MHD is turned on if we're trying to use anisotropic conduction.
   // if not, alert user.
   if(AnisotropicConduction==TRUE && UseMHD==0){
-    ENZO_FAIL("AnisotropicConduction can only be used if MHD is turnedf on!\n");
+    ENZO_FAIL("AnisotropicConduction can only be used if MHD is turned on!\n");
   }
   if(AnisotropicConduction==TRUE && MetaData.TopGridRank < 2){
     ENZO_FAIL("AnisotropicConduction can only be used if TopGridRank is >= 2!\n");
--- a/src/enzo/ReadParameterFile.C
+++ b/src/enzo/ReadParameterFile.C
@@ -78,7 +78,54 @@ int ReadParameterFile(FILE *fptr, TopGridData &MetaData, float *Initialdt)
   char **active_particle_types;
   active_particle_types = new char*[MAX_ACTIVE_PARTICLE_TYPES];
   int active_particles = 0;
-  
+
+
+#ifdef USE_GRACKLE
+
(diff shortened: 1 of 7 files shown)


brittonsmith:
@aemerick, this looks like a good solution, but it adds quite a few lines to `ReadParameterFile.C` that may seem a bit counter-intuitive to someone checking out that code. Can I suggest moving all of the Grackle parameter setting/parsing into a separate file like how `CosmologyReadParameters.C` operates?

aemerick:
@brittonsmith That's not a bad idea. If I'm going to make that change, should I just make a `GrackleReadParameterFile.C` and move reading every Grackle related parameter from `ReadParameterFile.C` into there?

aemerick:
@brittonsmith  I wen't ahead and assumed the answer was "yes"

gregbryan:
I very much like the idea of moving all the grackle related things into its own file!  However, I may be missing something, but it seems to me this particular way of doing it changes the behaviour.  I *think* the old way worked as follows to deal with parameters that were both in Enzo and grackle: (i) it initially set the Enzo defaults, (ii) then used the grackle defaults to set the Enzo parameters, (iii) then allowed the Enzo parameters to be overridden by the parameter file, and (iv) finally copied the Enzo parameters back over the grackle parameters so that only those Enzo/grackle parameter specifically changed in the parameter file were changed in grackle.  Otherwise, they had the Grackle default.  The new way seems different -- it does (i) (iii) (ii) and (iv) which means that the parameter file won't change the enzo parameters.  For example, this new way will always have Gamma set equal to the grackle default.  As I said, it is late at night, so I might be missing something obvious...

aemerick:
@gregbryan I think you're right. This does change the behavior. 

I think this means we either go back to before the latest commit where everything is in `ReadParameterFile.C` (I'm not too opposed to leaving everything there). 

Or if we want to have everything in one file then the two options for this are either: A) move all heating / cooling / chemistry parameters to the same file (not my preferred solution), or B) make a separate function (e.g. `GrackleSetDefaultParameters`) to do (ii) which would be called right before the large parameter read block in `ReadParameterFile`, then goes through this block to do (iii), then calls `GrackleReadParameters` to do (iv).  (and included in (iv) is reading Grackle specific parameters).

src/enzo/ReadParameterFile.C
```diff
@@ -1647,8 +1622,8 @@ int ReadParameterFile(FILE *fptr, TopGridData &MetaData, float *Initialdt)
   // make sure that MHD is turned on if we're trying to use anisotropic conduction.
   // if not, alert user.
   if(AnisotropicConduction==TRUE && UseMHD==0){
-    ENZO_FAIL("AnisotropicConduction can only be used if MHD is turned on!\n");
-  }  
+    ENZO_FAIL("AnisotropicConduction can only be used if MHD is turnedf on!\n");
```
brittonsmith:
```suggestion
    ENZO_FAIL("AnisotropicConduction can only be used if MHD is turned on!\n");
```

gregbryan:
I also like plan B!

aemerick:
@brittonsmith Yep, thats exactly it. Just made the change.

aemerick:
Also, apologies for the large diff in `ReadParameterFile.C`... apparently my text editor is really into removing white space at the end of lines

src/enzo/ReadParameterFile.C
```diff
@@ -78,29 +80,38 @@ int ReadParameterFile(FILE *fptr, TopGridData &MetaData, float *Initialdt)
   char **active_particle_types;
   active_particle_types = new char*[MAX_ACTIVE_PARTICLE_TYPES];
   int active_particles = 0;
-  
+
+#ifdef USE_GRACKLE
```
brittonsmith: I think this `ifdef` needs to be removed because there is the code inside `GrackleSetDefaultParameters` that checks for `use_grackle = 1` and grackle not compiled.

src/enzo/ReadParameterFile.C
```diff
@@ -1739,94 +1724,23 @@ int ReadParameterFile(FILE *fptr, TopGridData &MetaData, float *Initialdt)
 
 
 #ifdef USE_GRACKLE
```
brittonsmith: I think this `ifdef` can be removed for similar reasons since there is `ifdef` code inside `GrackleReadParameters`. This should allow you to remove all grackle ifdefs from this file, which should make it a bit easier to read.

src/enzo/ReadParameterFile.C
```diff
@@ -1862,30 +1776,30 @@ int ReadParameterFile(FILE *fptr, TopGridData &MetaData, float *Initialdt)
...
+      RadiationData.RadiationShield = FALSE;
 
     if ((RadiationFieldType >= 10 && RadiationFieldType <= 11) ||
         RadiationData.RadiationShield == TRUE)
       if (InitializeRadiationFieldData(MetaData.Time) == FAIL) {
 	ENZO_FAIL("Error in InitializeRadiationFieldData.");
       }
- 
+
 #ifdef USE_GRACKLE
```
brittonsmith: this one, too.

src/enzo/ReadParameterFile.C
gregbryan: Agreed -- good catch!

jwise77:
All of these changes look good to me, and I think this is a good solution to the problem we were having before. Approving!

PR #133, opened 2019-12-12
Bugfix for ActiveParticleFinalize function call

DeovratPrasad:
ActiveParticleFinalize function removed from the loop over grids.  Active particle counting over all grids and processors is already done 
inside CommunicationUpdateActiveParticleCount function which is getting called in ActiveParticleFinalize function. So, no need to loop over grids again.


jwise77:
We had a bug recently introduced that made the tests fail when you submitted your PR. It was fixed yesterday in #131. Could you merge the latest changes from enzo-project:master so the tests can run again?

gregbryan:
@DeovratPrasad - if you can merge the latest changes from enzo-project:master as suggested by John, we can confirm that the tests pass and merge this.  Thanks again!

DeovratPrasad:
@gregbryan Done. Thanks!!

PR #134, opened 2020-01-02
Updating gold standard

jwise77:


jwise77:
The initial test failed because I think I forgot to push the new `gold-standard-v14` tag. I'm rerunning the tests after explicitly pushing the tags.

jwise77:
Tests pass after pushing tag.

PR #141, opened 2020-04-03
Minor changes for StochasticForcing

jcforbes:
This commit fixes two issues my collaborators and I have encountered when using the StochasticForcing problem:

1) The docs have an error for the DrivenFlowProfile parameter, in which the value for 'band' and 'parabolic window' are switched. This conflicts with correct statements in the comments of the .enzo file  run/MHD/3D/StochasticForcing/StochasticForcing.enzo and the definitions in typedefs.h from lines 201-204.

2) For certain sets of compilers, some amplitudes of the driving field are incorrectly non-zero, because they are never explicitly set in the code. Here I just add some else statements that guarantee these amplitudes are zero when they should be in the Band case.

### DIFF
--- a/doc/manual/source/parameters/index.rst
+++ b/doc/manual/source/parameters/index.rst
@@ -5020,7 +5020,7 @@ Turbulence Simulation with Stochastic Forcing (59)
 
 
 ``DrivenFlowProfile`` (external)
-    Shape of forcing power spectrum (1: delta peak, 2: band, 3: parabolic window).
+    Shape of forcing power spectrum (1: delta peak, 2: parabolic window, 3: band).
 
 ``DrivenFlowAlpha`` (external)
     Ratio of domain length to integral length for each dimension (L = X/alpha).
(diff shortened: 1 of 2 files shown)


jwise77:
@pgrete Could you review this PR? Thanks!

pgrete:
I'll take care of this by the end of the week.

jwise77:
Hi @jcforbes, could you address @pgrete comments? Then I'll merge.

jcforbes:
Hi @jwise77 thanks for the reminder! 

Obviously @pgrete 's first comment is straightforward to address, and I've changed that in my local copy. For the second comment, I haven't been able to verify that replacing my changes with the initializer solves (or doesn't solve) the issue. Not sure whether my test case is not quite the same as what I was using before, or if something has changed on the system I'm using. For now, as long as everyone's ok with it, I'll just keep the admittedly inelegant else statements.

jwise77:
3 approves. I'm merging!

PR #145, opened 2020-04-20
Star particle class feedback fixes and WIP Fire-2 stars in Enzo

azton:
This document describes the changes to ```Enzo``` from a) fixing Pop3 and stellar cluster feedback deposition, and b) implementing FIRE-2 like stars in ```Enzo```

# Pop 3/Cluster fixes

## The Problem:

In a study of metal evolution in the Renaissance Simulations, we discovered that there were massive errors in the metals deposited by Pop 3 supernovae (SNe, HNe, and PISN).  This error was resolution dependent, and in large part due to depositing a density calculated on a sphere to a grid structure:
DxScalingP3Feedback

This shows a constant density being deposited on a grid of varying resolution--but the center of the sphere is varied within the central cell (along diagonal i=j=k).  As resolution approaches 50 pc, the error in deposition is nearly 50x.   Of course, the renaissance simulations generally had higher resolution in star-forming cells, so at least two processes could lead to a higher error as we see here.  

* The star could move to a lower density region before supernova, thereby having a lower resolution at the most critical time. 
* When deposition occurred, it is processed at the smallest possible grid, but then is also deposited on higher grids for a consistent solution up the levels of grids--this made it so that the correct high-resolution deposition would be incorrectly added at coarser grid levels.

As we can see, the error in the Renaissance Simulations is massive (this is the Rarepeak_LWB simulation):

P3_RS_RP_LWB_errors

The spike in error at the end is an artifact of jumping from RD0008 $\to$ RD0041.  The error is taken as metal measured on the grid divided by the theoretical yeilds in Wise, et al., 2012.  Errors are $> 10^3$.  (as point of reference for spatial resolution, the Rarepeak_LWB has $dx_{min} = 19.6$ pccm at AMR level 12).  

The problem is less severe, but still apparent in higher resolution simulations, i.e., *Birth of a Galaxy*, where we found 2-3x too much Pop 3 metals on the grid:

p3_256-sv3

## The solution:

I implemented 3 fixes to alleviate these problems.  First, the density to deposit ($\rho_d$) to the grid is taken as the mass of ejecta in a sphere: $\rho_{d} = \frac{M_{ej}}{V_{sphere}}$.  Afterward, that density is taken at truth and deposted across the grid.  To make this more consistent, I added a routine ```Star_CalculateFeedbackVolume.C``` which simply sums up the volume that will be affected by the deposition.  Unfortunately, the "sphere" may deposit across grids, and hence across processes/tasks.  In order to get the true sum of all volume affected, we have to sum the affected volume across processes, which I do with a all-to-all mpi communication.  

The density is then rescaled $\rho_d = \frac{M_{ej}}{V_{sphere}}\times \frac{V_{sphere}}{V_{grid}}$.  The same rescaling factors are applied to metal density and energy density.  In a single deposition, with varying resolution (by setting must refine resolution, as below), we see very low errors (0.0 is perfect here):

SingleEvent_P3_fixed

The result is robust across thousands of Pop 3 depositions in a Renaissance-esque simulation with parameters and mass/spatial resolution matching the RS but in a cosmological AMR-everywhere box (1.0 is perfect here):

MetalErrorFixP3

The exact same routines are called for Pop 2 stellar cluster feedback depostion, so that has been fixed with this as well.  One remaining concern is that the blocking MPI communications may be prohibitive as the number of Pop 2 clusters gets large.  

In addition to the algorithmic fixes, I also wanted to ensure that ```Enzo``` users could be guaranteed that the thermal deposition from Pop3 stars was robust with their choice of resolution.  Toward that end, I added a new refinement criteria (option ```20```) where the code will refine a Pop 3 star to ```PopIIISupernovaMustRefineResolution``` cells in diameter for the stars lifetime + ```PopIIIMustRefineRegionLifetime``` Myr afterward.  This ensures that Pop III feedback occurs at the intended resolution, and that PopIII supernovae have a chance to evolve in that higher resolution region.  The above plot of Single SN error does not use this feature, but the new renaissance simulation showing thousands of events does use this parameter set with ```PopIIISupernovaMustRefineResolution = 4``` and ```PopIIIMustRefineRegionLifetime = 3```.

# FIRE-2 Stars (WIP)

This project is currently unfinished (ie., requires testing), but accidentally got rolled into the same branch with Pop2/3 fixes.  The code I've put in attempts to implements FIRE-2 style stellar formation and feedback routines in Enzo.  This is still in progress, so I have not added anything to the docs, as I don't want anyone to think it is a finished project and good to use for science (yet).  I opted to write this in C (though there is an outdated fortran routine as well).  All the major routines specific to this method can be found in \*MechStars\*.C files.

## Features

Star formation:

​	Star formation follows most of the standard Cen & Ostriker formulation, with additional virial parameter requirements as outlined in Hopkins, et al. 2018. 

### Formation Criteria: in MechStars_checkCreationCriteria.C

* $\rho > \rho_{crit}$
* $\nabla\cdot v < 0$ (converging flow)
* virial parameter $\alpha < 1$
* $T_{cool} < T_{dyn}$ (cooling time < freefall time)
* $f_{shld} > 0$ (Hydrogen shielding some fraction of gas)
* $M_{B} > M_{min}$ (mass in cell > min mass of star)
* OPTIONAL:
  * $Z > Z_{crit}$
  * refinement level $l > l_{min}$

For the optional criterion, if $Z < Z_{crit}$, there is an option to include coarse PopIII seed fields, where we pull a Pop III mass from the IMF (as in the Star Particle class Pop 3 routines), and set off a Pop 3 supernova with the corresponding energy and metal deposition, using the mechanical stars feedback routine.  This will give a stochastic initial metallicity field, but without the computational complexity of following Pop III evolution explicitly.

### Feedback in Grid_MechStarsFeedbackRoutine.C and Grid_MechStarsDepositFeedback.C:

* Supernova mass, energy, metal yeild are calculated as in Hopkins, 2018

* Ala Simpson, 2015, we create a cloud of virtual coupling particles at r = dx from the source

* Each coupling particle is assigned a fraction of the deposition according to its radius from the source, mean density, and mean metallicity.

* coupling particles are deposited to the grid in CIC fasion.

* Can enable ionizing radiation (with rates from Hopkins 2018), using the Star Particle class ability to use standard star particles as ionizing sources.

* Instead of coupling $\Delta \vec p = \min(\delta\vec p_{ej}, \delta \vec p_{terminal})\times\min[\sqrt{1+m_b/m_{ej}}, p_t/p_ej]$ as in Hopkins, we calculate the phase of the SN based on, ie., density, energy, metallicity, and couple one of $p_{free}$ for free-expansion, $p_{sedov}$ for Sedov-Taylor, $p_{shell}$ for shell-forming stage, or finally $p_t$ for terminal phase.

* Each star particle can undergo continuing accretion based on the same criteria as star formation, but without a minimum formation mass (can add arbitrarily small amounts of mass).

  * Only allowed for first 50 Myr -- after that, normal creation is preferred to get more SNe generating star particles.
  * Metals fraction of particle is mass-weighted during continuing formation

## Parameters

Bold parameters are defaults.

* `StellarWinds` = {0,**1**}
  * Allow stellar winds from particles
* `SingleSN` = {0,**1**}
  * Allow probabilistic SNe (0 implies no SNe at all)
* `StarMakerMaximumFormationMass` = FLOAT
  * Maximum mass that can be formed from a single cell. Prevents total evacuation of cells that are very dense
* `StarMakerMaximumMass` = FLOAT
  * Absolute maximum mass allowed for a star particle
* `DepositUnresolvedEnergyAsThermal` = {**0**,1}
  * if `1`, will do NGP deposition for winds to reduce computational load (for many stars, CIC can be too slow to be reasonable)
* `NEvents` = INT (**-1**)
  * if -1, will probabilistically model supernova.  If INT > 1, will set off N-supernovae per particle, (1 per timestep). Mostly for testing purposes
* `AnalyticSNRShellMass` = {**0**, 1}
  * on- increases shell mass of SNR based on analytic formulae (Not tested)
* `StarMakerMinimumRefinementLevel` = INT (**0**)
  * Wont allow star formation at coarser levels than this.
* `UnrestrictedSN` = {**0**, 1}
  * on - allows any number of SNe per particle per timestep. off-only allows one SNe per timestep, and limits timesteps according to `StarMakerMaximumMass` to ensure that a maximum mass particle will only have $\leq 1$ SN per timestep.
* `MechStarsCriticalMetallicity` = FLOAT 
  * Works in conjunction with `MechStarsSeedField`, if both set, then wont form stars below this metallicity, and will instead form a Pop III seed supernova if all formation criteria (except metallicity) are met.  
* `MechStarsRadiationSpectrum`={**0**, 1}
  * Whether to couple stars to radiation via ray-tracing  (in the Star Particle class method)

* `MechStarsSeedField` = {**0**, 1}
  * Enables SN_Colour field, and sets off seed P3 supernovae before star formation can occur.  These seed SN must meet all formation criteria except metallicity.

### Test Star parameters

I added the mechanical formulation to the TestStarParticle problem, in addition to some new parameters

* `NumberOfTestStars` = INT (**1**)
  * Now, can have any number of stars in a cluster for testing
* `ClusterRadius` =  FLOAT (**0.0**)
  * Define the radius of the cluster, in fraction of box-width

## TODO

* General Cleanup
* More testing.  Have completed feedback tests at multiple scales, but in an initial run, found the feedback ineffective for the spatial resolution I require, and so moved on to other work.  Need another test at high resolution, $\sim$ FIRE-2 resolution, to validate the formation and feedback routines work as intended at high-resolution.  In particular, this seems to create too many stars at high redshift (a $10^{10} M_\odot$ halo generates $\sim 10^8 M_\odot$ stars by $z = 9$ with $dx\sim 40$ pccm.)

### Some Plots showing the routine in action

Single SN scaling momentum deposition $\sim 100 $ kyr post SN:

MechStars_P_Scaling

A Single SN a few timesteps after deposition:
MechStars_singleSN

A cluster of 20 1k $M_\odot$ stars going through the main sequence after 36 Myr

data0150_Projection_z_density_density


aemerick:
Hey @azton, Glad to see this PR! 

I know it may be a pain to do so, but if the Pop 3 / StarParticleClass bugfix is ready to go and not reliant on the FIRE additions which are WIP, would it be worth spinning that portion off as a separate PR since it is a fairly significant bugfix. (Unless you have a well-defined timeline for when the FIRE stuff will be completed). Not sure if others (@jwise77 ?) have thoughts on this...

azton:
@aemerick I thought about doing that, but Mike told me @jwise77 wanted to use the Pop 3 bugfix on Frontera ASAP, and to put out the PR as is to get it available as soon as possible.   I don't have a timeline when I'll be able to finish testing on the FIRE2 stuff, as I'm coming up short on my PhD timeline and its on the back burner.

aemerick:
@azton Ah ok. That makes total sense. And ya prioritize your PhD over this. 

But @jwise77 if you are planning on using the bugfix for runs before the WIP is completed (which is sounds likely?) would you be likely to just pull from this PR directly or just manually implement the bugfix changes to your fork?

jwise77:
I was going to pull from this PR, test it, and use it directly. I don't want to manually implement the changes because it'll create conflict headaches down the line.

fearmayo:
Hi All, 

Where are we at with this PR? Is it ready for a full review?

azton:
@fearmayo The star particle class fixes should be ready to review--I expect I'll have to make some changes to fit in with Enzo's best practices, etc, but the simulations I'm running with these fixes are progressing well.  I wanted to get the FIRE-2 style feedback out for the community, but I don't have the computer or physical time to finish it currently (I'm working on the last leg of my PhD, so that project has been put onto the far back burner).  I want to come back to it eventually, but that is dependent on a lot of other factors right now.

fearmayo:
Hi Azton, 

Just an FYI - it looks like your build fails in several of the Grid_MechStars* files when the precision is changed i.e. when the definition of FLOAT changes. Have a look here: https://enzo.readthedocs.io/en/latest/developer_guide/FloatIsDouble.html

jwise77:
After a discussion in the Enzo workshop, we decided to break off the metal deposition bug fix and have Will and Claire issue a new PR.

PR #120, opened 2019-09-25
Star Particle Communication Bugfixes

aemerick:
This fixes / improves star particle communication to address memory fragmentation and performance issues when running with many (> 10k) star particle objects.

These changes are all circa 2016, but fell through the cracks in getting pulled into the main repository. Caught this when merging my fork with the active particle changes. Some of these changes parallel what is done with active particles. 

All credit goes to John Wise, stemming from the following commits: [May 2016 (1 of 2)](https://bitbucket.org/jwise77/enzo-dev/commits/ba5202af3d2c6102bd82ca644fed3c373e6552e8), [May 2016 (2 of 2)](https://bitbucket.org/jwise77/enzo-dev/commits/703f1091d58a4e12cc6720bc08ef12581e56792f), [June, 2016](https://bitbucket.org/jwise77/enzo-dev/commits/2981eb07b9effb06e490a38047fefa5bfb8237a2)

PR #150, opened 2020-05-20
Bugfix: must-refine particles identification by mass

jwise77:
**Star objects:** The recent change in #120 was missing a line that first checks whether there are any local star objects before copying them to a comm buffer.

**Must-refine:** Allow for some machine-precision epsilon when comparing particle masses to determine which initial nested level they came from. Also related, the `f` suffixes in the EPSILON defines caused them to be 32-bit floats instead of 64- or 128-bit.

### DIFF
--- a/src/enzo/GrackleReadParameters.C
+++ b/src/enzo/GrackleReadParameters.C
@@ -218,7 +218,8 @@ int GrackleReadParameters(FILE *fptr, FLOAT InitTime)
   code_units grackle_units;
   grackle_units.a_units = 1.0;
   float DensityUnits = 1.0, LengthUnits = 1.0, TemperatureUnits = 1.0,
-    TimeUnits = 1.0, VelocityUnits = 1.0, MassUnits = 1.0;
+    TimeUnits = 1.0, VelocityUnits = 1.0;
+  double MassUnits = 1.0;
   if (GetUnits(&DensityUnits, &LengthUnits, &TemperatureUnits,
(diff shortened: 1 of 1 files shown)


jwise77:
I forgot that @aemerick had already included the star particle fix in #140. Hah!

jwise77:
The PhotonTest is failing. I will look into this.

aemerick:
Happy to take a look at this once the photon test thing gets sorted out. I'm a bit surprised the must refine machinery has anything to do with that. While you're at it, could you fix a minor memory leak I just found (at least, on some compilers). On line 99 of `Grid_DepositMustRefineParticles.C` it has 

`antirules = new bool[NumberOfAntiRules];`

when `NumberOfAntiRules` is initially 0, and only de-allocates this space if `NumberOfAntiRules > 0`. Which apparently needs to be done even if its `new bool [0]`.... Safer to just initialize this pointer to `NULL`.

gregbryan:
It looks like you are using gold-standard-v13 rather than gold-standard-v14, which I think has this failure in generating the answers.  Maybe you need to merge against the latest repo?

gregbryan:
Otherwise, the code looks good to me!

gregbryan:
> It looks like you are using gold-standard-v13 rather than gold-standard-v14, which I think has this failure in generating the answers. Maybe you need to merge against the latest repo?

jwise77:
> But. While you're at it, could you fix a minor memory leak I just found (at least, on some compilers). On line 99 of `Grid_DepositMustRefineParticles.C` it has

Done!

Issue #142, opened 2020-04-03
Restarting with SelfGravity fails when WritePotential = 1


jcforbes:
Steps to reproduce:

Go to run/MHD/3D/StochasticForcing
Run the problem to produce at least one snapshot.
Modify the restart parameter file DDXXXX/dataXXXX to set SelfGravity = 1 and WritePotential = 1
Restart the run from that snapshot.

Result:
```
Caught signal 11 (Segmentation fault: address not mapped to object at address 0xd)
==== backtrace ====
    0  /lib64/libucs.so.0(+0x18fbc) [0x2b324db6bfbc]
    1  /lib64/libucs.so.0(+0x19172) [0x2b324db6c172]
    2  ./enzo.exe() [0x636219]
    3  ./enzo.exe() [0x48dc0d]
    4  ./enzo.exe() [0x48cc32]
    5  ./enzo.exe() [0x77abb4]
    6  ./enzo.exe() [0x4ddeed]
    7  ./enzo.exe() [0x4ce90e]
    8  ./enzo.exe() [0x40e5df]
    9  /usr/lib64/libc.so.6(__libc_start_main+0xf5) [0x2b324c8ba3d5]
   10  ./enzo.exe() [0x40d889]
===================
Segmentation fault (core dumped)
```

For now the solution is just to set WritePotential = 0, so maybe the solution is as easy as adding a note in the docs somewhere, though I guess in principle in an ideal world one should be able to have the potential be written out.  I did do some digging, and the proximate cause of the segfault is this line:
` BaryonField[field][jj++] = PotentialField[index];`
of Grid_CopyPotentialToBaryonField.C, but I don't understand the parts of the code leading up to that, involved in reading in restarts and setting the number of baryon fields, well enough to understand why the value of WritePotentialField matters.

aemerick:
I *think* what is happening is that when both SelfGravity and WritePotential are on during a restart, Enzo tries to read the potential from the restart dump. If WritePotential is initially off and these are turned on midway through a run, the potential field doesn't exist in the output and Enzo seg faults trying to read it. However, it looks like Enzo does try and do some error checking in `Grid_CopyPotentialToBaryonField.C` that should catch this, but it does not for some reason.

The fix here would be to have Enzo do a better job checking if the potential field exists on disk, and then computing it if it does not (rather than trying to copy it). 

For now, you can have Enzo compute the potential and write it to disk for you, at which point you should be able to restart safely. This can be done with the "-g" flag, or:

[...]

PR #151, opened 2020-05-21
Restarting with changes to grav potential parameters (plus minor changes)

jwise77:
This bugfix addresses #142 and allows for `SelfGravity` to be turned on from a restart.

When I was implementing this fix, I also included a few other minor changes by mistake, and I didn't feel it was necessary to separate them now.
1. Bugfix to the metal initialization in CollapseTest
2. Put the center of mass stdout of CollapseTest in the `Debug1` setting to reduce verbosity
3. Allow CollapseTest to have radiating sources
4. Bugfix to "accretion sphere" after star formation to keep the metallicity unchanged


src/enzo/New_Grid_ReadGrid.C
```diff
@@ -223,10 +224,18 @@ int grid::Group_ReadGrid(FILE *fptr, int GridID, HDF5_hid_t file_id,
  
     /* 4) Read gravity info */
  
-    if (SelfGravity)
+    if (SelfGravity) {
       if (fscanf(fptr, "GravityBoundaryType = %"ISYM"\n",&GravityBoundaryType) != 1) {
-		ENZO_FAIL("Error reading GravityBoundaryType.");
+	// If turning on self gravity in a restart, then determine
```
gregbryan, line 229: This is OK, but after double-checking the code, one slightly cleaner way to do this would be to remove the check here and just set the GravityBoundaryType in Group_ReadDataHierarchy.  You do this anyway for the topgrid, but you could also include subgrids in that by checking to see if ParentGrid != NULL and then set GravityBoundaryType = SubGridIsolated (otherwise set to MetaData.GravityBoundary as you do now). That would remove this slightly awkward calculation here and push the check into the right level. In that case, the error check for reading GravityBoundaryType could just silently fail. This would also have the advantage that the code would work if NEW_GRID_IO were turned off (right now I think it would not since you don't have this bit of code in the old grid reader, Grid_Group_ReadGrid.C).

src/enzo/Group_ReadDataHierarchy.C
```diff
@@ -225,6 +225,10 @@ int Group_ReadDataHierarchy(FILE *fptr, hid_t Hfile_id, HierarchyEntry *Grid,
                                      MetaData.PPMFlatteningParameter,
                                      MetaData.PPMDiffusionParameter,
                                      MetaData.PPMSteepeningParameter);
+  if (SelfGravity) {
```
gregbryan, line 228: As noted in another comment, this could include subgrids by checking ParentGrid (to see if it is NULL).

jwise77:
I only tested this in serial and it fails in parallel on restart. I will look into this further along with @gregbryan comments.

src/enzo/Grid_AddFeedbackSphere.C
```diff
@@ -922,10 +922,10 @@ int grid::AddFeedbackSphere(Star *cstar, int level, float radius, float DensityU
 	    }
 
 	    if (SNColourNum > 0)
-	      BaryonField[SNColourNum][index] *= factor;
```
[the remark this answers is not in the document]
jwise77: Thanks for catching this. After a star particle is formed, the "accretion sphere" is replaced with a uniform density. The metal field was only reduced by that factor (0.57), but the more correct thing to do this is to keep the same metallicity as we change the density.

I'll add a note in the PR description.

src/enzo/New_Grid_ReadGrid.C
jwise77: I like this approach a lot better. I've made your suggested change.

jwise77:
I've now addressed the comments from @gregbryan.

src/enzo/CallProblemSpecificRoutines.C
```diff
@@ -42,7 +42,7 @@ int CallProblemSpecificRoutines(TopGridData * MetaData, HierarchyEntry *ThisGrid
     ThisGrid->GridData->SphericalInfallGetProfile(level, 1);
   if (ProblemType == 30)
     ThisGrid->GridData->AnalyzeTrackPeaks(level, 0);
-  if (ProblemType == 27){
+  if (ProblemType == 27 && debug1){
```
brittonsmith, line 45: I seem to recall that `debug1` results in a lot of output. Would `debug` be better?

src/enzo/Grid_AddFeedbackSphere.C
```diff
@@ -922,10 +922,10 @@ int grid::AddFeedbackSphere(Star *cstar, int level, float radius, float DensityU
...
+	      BaryonField[SNColourNum][index] = metallicity * BaryonField[DensNum][index];
```
brittonsmith, line 925: Won't this set the `SNColourNum` field to the same value as the `MetalNum` field? Is that the right thing to do?

src/enzo/New_Grid_ReadGrid.C
```diff
@@ -386,55 +387,58 @@ int grid::Group_ReadGrid(FILE *fptr, int GridID, HDF5_hid_t file_id,
...
-      
-    } /* if HydroMethod == MHD */
-
-
     delete [] temp;
  
   }  // end:   if (NumberOfBaryonFields > 0 && ReadData &&
   //      (MyProcessorNumber == ProcessorNumber)) {
 
+  if (HydroMethod == MHD_RK) { // This is the MHD with Dedner divergence cleaning that needs an extra field
```
brittonsmith, line 395: Did this block of code move in a critical way?

src/enzo/CallProblemSpecificRoutines.C
jwise77: I agree that `debug` would be better.

jwise77: I agree that `debug` would be better.

src/enzo/Grid_AddFeedbackSphere.C
jwise77: Good catch. They should be the same fields. In principle, it should happen for all of the metal fields. I'll expand it to include them.

src/enzo/New_Grid_ReadGrid.C
jwise77: Yes, it did. See this [commit message](https://github.com/enzo-project/enzo-dev/pull/151/commits/34dfe0f1523149277003061d28dc11e98689458e). Before it was in the `ProcessorNumber == MyProcessorNumber` if-block, so it was only added on a single MPI process. This only worked in serial restarts before and crashed in parallel restarts. The `AddField()` routine adds the new field to the `FieldList` and only allocates and initializes the array on the "owner" MPI process.

jwise77:
@brittonsmith I finally got around to answering your questions after 2 years!

jwise77:
@brittonsmith Can you look over the latest changes and approve? Thanks!

brittonsmith (approved):

PR #153, opened 2020-06-02
Memory leak fix - deleting radiation sources 

aemerick:
Memory leak fix in radiation sources:

1) small fix in `EvolvePhotons`. Sources would sometimes (all of the time?) not get deleted properly in the main EvolvePhotons loop after the end of their life. Placing an additional check after the loop to see if there are any sources with lifetimes < PhotonTime. This is likely not that large of a leak since it only occurs once per radiation source (at the end of its life) and the `RadiationSourceEntry` structs aren't very large. 

2) Larger leak fix. `StarParticleRadTransfer.C` deletes and re-creates the `GlobalRadiationSources` list every call, but this was not being done correctly. Using call to `DeleteRadiationSource` properly now which properly deletes the allocated arrays in the struct.

### DIFF
--- a/src/enzo/EvolvePhotons.C
+++ b/src/enzo/EvolvePhotons.C
@@ -214,6 +214,9 @@ int EvolvePhotons(TopGridData *MetaData, LevelHierarchyEntry *LevelArray[],
                        MAIN RADIATION TRANSPORT LOOP
    **********************************************************************/
 
+  int NumberOfSources = 0;
+  bool DeleteSources = FALSE;
+
   while (GridTime > PhotonTime) {
 
     /* Recalculate timestep if this isn't the first loop.  We already
@@ -236,8 +239,8 @@ int EvolvePhotons(TopGridData *MetaData, LevelHierarchyEntry *LevelArray[],
     RS = GlobalRadiationSources->NextSource;
     LastNode = RS_GridList;
     TempGridList = RS_GridList->NextGrid;
-    int NumberOfSources = 0;
-    bool DeleteSources = (FirstTime ||
+    NumberOfSources = 0;
+    DeleteSources = (FirstTime ||
 			  !(RadiativeTransferAdaptiveTimestep == FALSE &&
 			    RadiativeTransferSourceClustering == TRUE));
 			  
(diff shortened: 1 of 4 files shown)


aemerick:
I should add, this does not fix my seg fault issue that I e-mailed the enzo-dev list about last week.

src/enzo/EvolvePhotons.C
```diff
@@ -716,6 +719,28 @@ int EvolvePhotons(TopGridData *MetaData, LevelHierarchyEntry *LevelArray[],
...
+  /* Double check and clean up photon sources that have expired */
+
+  if (NumberOfSources > 0){
+    RS = GlobalRadiationSources->NextSource;
+
+    /* Clean up photon sources if time is up */
+    while (RS != NULL){
+      if ( ((RS->CreationTime + RS->LifeTime) < PhotonTime) && LoopTime == TRUE &&
+	           DeleteSources) {
+      if (debug) {
```
brittonsmith, line 731: This is a real nitpick, but the tabbing here and with the print statements is a bit inconsistent. Could you line those up?

src/enzo/StarParticleRadTransfer.C
```diff
@@ -51,10 +53,13 @@ int StarParticleRadTransfer(LevelHierarchyEntry *LevelArray[], int level,
   /* If sources exist, delete them */
 
   RadiationSourceEntry *dummy;
-  while (GlobalRadiationSources != NULL) {
-    dummy = GlobalRadiationSources;
-    GlobalRadiationSources = GlobalRadiationSources->NextSource;
-    delete dummy;
+  if (GlobalRadiationSources!=NULL){
```
brittonsmith, line 56: I think this is the same block as in `ActiveParticleInitialize.C` Maybe this could be a separate function like `DeleteAllRadiationSources`?

src/enzo/EvolvePhotons.C
aemerick: Done!

src/enzo/StarParticleRadTransfer.C
aemerick: Done!

jwise77:
Three approves. I'm merging!

PR #156, opened 2020-08-12
fix ci testing

brittonsmith:
The latest major release of matplotlib clashes with the lastest stable yt release, which is why the tests are failing. This avoids installing the clashing matplotlib version. I also went ahead and updated to the latest version of Python for the testing.

The last commit adds names to the circleci script so that the appearance on circleci shows that name and not a smashed together string of all the commands. Unfortunately, it requires everything to go two spaces to the right, but it does make the ci easier to parse.

Apparently the docs build was also failing with a dead intersphinx link, which led me to discover a number of links to yt.enzotools.org, which I have changed to yt-project.org. Unfortunately, we also linked to a script that is no longer available, so I just left a note about that.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,7 +26,8 @@ commands:
           pip install --upgrade pip
           pip install --upgrade wheel
           pip install --upgrade setuptools
-          pip install Cython numpy h5py matplotlib fastcache flake8 nose yt girder-client
+          pip install 'matplotlib<3.3.0'
+          pip install Cython numpy h5py fastcache flake8 nose yt girder-client
           # Install hypre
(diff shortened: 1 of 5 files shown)


fearmayo:
Looks great! How sustainable is this though? It looks like you force an older version of matplotlib to be pulled. What happens in the future when this is no longer viable?

brittonsmith:
> Looks great! How sustainable is this though? It looks like you force an older version of matplotlib to be pulled. What happens in the future when this is no longer viable?

The development version of yt fixes this compatibility issue. When that is released we won't have to worry about this.

jwise77:
Three approves. I'm merging.

PR #157, opened 2020-09-01
add machine file for tigercpu at Princeton

HerculesJack:


jwise77 (approved):

src/enzo/Make.mach.tigercpu
```diff
@@ -0,0 +1,146 @@
...
+#               cd ../ring/
+#               make
+
+#-----------------------------------------------------------------------
+# Install paths (local variables)
+#-----------------------------------------------------------------------
+
+LOCAL_COMPILER = ${PROD_DIR}/compiler
+LOCAL_MPI_INSTALL = ${MPI_HOME}
+LOCAL_HDF5_INSTALL = ${HDF5DIR}
```
clairekope, line 40: If any combination of these are defined by the user (e.g. in the `.bashrc`), and not when a module is loaded, I would suggest putting the full path here, substituting system variables like `$USER` as appropriate.

clairekope (changes requested):
This file looks good, but I want confirmation that file paths such as `$HDF5DIR` will be available to any Tiger user and aren't specific to He Jia's shell environment.

src/enzo/Make.mach.tigercpu
HerculesJack: Hi, I believe those paths are defined when the corresponding modules are loaded.

clairekope: Excellent! I'll remove the change request and approve.

clairekope (approved):

src/enzo/Make.mach.tigercpu
peeples: Why all the mercurial? Neither enzo nor yt use anymore. (And yt should not be required for installing enzo.)

PR #165, opened 2021-02-26
Exit with error if metal cooling on but no metal field present.

brittonsmith:
This changes the behavior when metal cooling has been enabled, but there is no metal field present. Previously, we disable metal cooling and keep running. I argue this is not what the user would want. Personally, I would rather the simulation not run instead of running in a way I hadn't intended, especially since the warning statement will most likely be missed. As well, there is a bug here because while `MetalCooling` is set to 0, the corresponding parameter within Grackle is not, so this results in wild behavior as Grackle tries to access `BaryonField[0]` (i.e., the gas density) as the metal density.


gregbryan (approved):

bwoshea (approved):
This makes a lot of sense to me.  Thanks!

jwise77 (approved):

PR #171, opened 2021-04-12
Fix HDF5 include path

michaelkuhn:
`Make.mach.linux-gnu` contains a typo in `LOCAL_INCLUDES_HDF5` that causes the HDF5 header not to be found.


jwise77 (approved):

pgrete (approved):
Good catch!

fearmayo (approved):

PR #172, opened 2021-04-14
Removing DebugField from MHD collapse problem

jwise77:
This was causing an I/O crash because the data label was never associated with it in the problem initializer. h/t Armin M. on the mailing list for pointing this out.

I'm also including a minor fix to the MassUnits to be double instead of float. This causes compiler issues when using mixed precision.

### DIFF
--- a/src/enzo/GrackleReadParameters.C
+++ b/src/enzo/GrackleReadParameters.C
@@ -218,7 +218,8 @@ int GrackleReadParameters(FILE *fptr, FLOAT InitTime)
   code_units grackle_units;
   grackle_units.a_units = 1.0;
   float DensityUnits = 1.0, LengthUnits = 1.0, TemperatureUnits = 1.0,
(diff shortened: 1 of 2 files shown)


jwise77:
I'm certain that the type change is causing the test failure. I will look into this.

gregbryan:
> I'm certain that the type change is causing the test failure. I will look into this.

It looks like this is a problem with tags (your branch already has the gold-standard-v14 tag - perhaps easiest to just remove it from your branch?).

jwise77:
> > I'm certain that the type change is causing the test failure. I will look into this.
> 
> It looks like this is a problem with tags (your branch already has the gold-standard-v14 tag - perhaps easiest to just remove it from your branch?).

Oh, good catch! That's much easier to update and check.

fearmayo (approved):

gregbryan (approved):

PR #179, opened 2021-09-01
Fixes to the testing infrastructure

gregbryan:
This PR makes three changes related to testing:

1. It changes circleci -> cimg in the CircleCI config.yml to switch from outdated Ubuntu images.
2. It fixes where get_yt_version is imported from in test (yt 4 no longer supports the old one).
3. It pins the version of yt used for testing to 3.6.1 (which seems to support both get_yt_version imports). This can be reverted once we update the gold standard version to 15.

These changes address some failures showing up in otherwise correct PRs.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -177,7 +177,7 @@ commands:
 jobs:
   test-suite:
     docker:
-      - image: circleci/python:3.8.5
+      - image: cimg/python:3.8.5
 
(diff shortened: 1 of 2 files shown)


brittonsmith (approved):
Thanks for taking care of this! These look good to me.

jwise77 (approved):

PR #185, opened 2022-02-23
Test suite fixes to accomodate CircleCI changes

gregbryan:
This PR contains a few tweaks to the CircleCI configuration file to resolve a recent issue that broke the test suite when the maximum runtime of the tests was reduced to 60 minutes.  Both the test-suite and compilation-suite exceeded this. This PR has three minor changes:
(i) remove amr_cosmology and RayleighTaylor_CT_Suppressed tests. These two tests were the two longest and accounted for a large fraction of time. Note that the air_cosmology_nested test is still run, so there is still amr_cosmology test coverage.
(ii) removed a few of the compilation combinations which are tested to bring the test time below 60 minutes. Hopefully the compilation variations tested are still wide enough to catch most issues.
(iii) fixed an issue with the answer-test caching in CircleCI -- this wasn't working correctly (the tests were always regenerated even if successfully cached).


jwise77 (approved):
Looks good. Thanks!

brittonsmith (approved):
This looks good to me, although would it be a good idea to test compilations of different values of integer and particle id precision rather than the same for each?

PR #188, opened 2022-05-09
Making Enzo Work with an arm64 architecture

emfisk:
These changes allow Enzo to be compiled and run on computers with an arm64 architecture, like Apple's M1 MacBooks. The code added checks to see if you have an arm64 architecture, and if so, uses "sse2neon.h" instead of "xmmintrin.h", which doesn't work on arm64. It should also still work on computers with an x86 architecture.


jwise77:
Thank you for your contribution! All looks good to me. The test suite isn't passing because it's timing out because we need to sort out some issues with the CI setup.

jwise77 (approved):

jwise77:
@emfisk I fixed the testing in #190, and I merged in these changes to run the test suite on this PR. But I couldn't push the tags to your PR. Could you issue the command `git fetch https://github.com/jwise77/enzo-dev --tags` and then `git push` to bring in the new gold-standard-v15 tag for testing?

jwise77:
For the other review, you can skip over all of the change related to the test suite. I wanted to merge the updated testing into the remaining PRs before waiting on the testing PR to be merged.

emfisk:
Is there anything else you need me to do?

jwise77:
Nothing else. Thanks for checking in! We're just waiting for someone to test the compilation on a M1 device.

brittonsmith:
I've just tried this out and can confirm that enzo both compiles and runs with these changes. I also see that it does not compile without them.

brittonsmith (approved):

PR #189, opened 2022-06-09
Corrected min/max reporting in performance.h

jwise77:
For some reason, the min/max functions returned bool types instead of the templated type T. After this change, performance.h now reports the correct min/max values for each performance timer.

There is only 1 effective commit in this PR with the first one (b833391) being already incorporated into another PR and the others being merges. Whoever merges this PR should rebase/squash to remove them.


clairekope (approved):

jwise77:
For the other TBD reviewer, the only changed file is `EnzoTiming.h`. I wanted to pull in #190 to run the test suite.

gregbryan (approved):

PR #190, opened 2022-07-05
Updating test suite to yt-4.x and concurrent testing

jwise77:
This PR updates all testing infrastructure to be compatible with yt-4.x, including
* fields as tuples
* removing calls to the halo mass function
* referring to `yt_astro_analysis` for `HaloCatalog` calls
* removing `yt.mods` calls
* replacing `na` with `np` (!)

I created a new gold standard (v15) because the old gold-standard cannot be compared to the new runs without some changes to the testing environment (e.g. different packages and dependencies). I manually confirmed on my local machine that the latest PR (#132) passed the push suite before making the new gold standard.

Another big change was using process pools in the `multiprocessing` python module to run the test simulations concurrently. This saves about 10 minutes in the push suite simulations. I couldn't get the testing to run in parallel without putting more effort into it. This could be addressed in a future PR. I increased the CircleCI resource from medium to large since we can run tests in parallel now, and the large nodes are still free to us.

Note: There are a lot of duplicate changes in common code in various files, so the review should be easier than what the diff size says.


brittonsmith (approved):
A heroic effort, John! Thanks for doing this. This looks good to me.

pgrete (approved):
Thanks for the fixing this. Looks good to me. I only have a couple of non-blocking clarification questions.

run/Hydro/Hydro-3D/RotatingSphere/profile_script.py
```diff
@@ -4,7 +4,8 @@
 mpl.use('Agg')
 import matplotlib.pyplot as plt
 import numpy as np
-from yt.mods import *
+import os
+import numpy as np
```
gregbryan, line 8:
```suggestion
```
This is a duplicate (already imported two lines earlier).

run/MHD/2D/MHD2DRotorTest/MHD2DRotorTest__test_almost_standard.py
```diff
@@ -3,7 +3,8 @@
 ###
 
 import os
-from yt.mods import *
+import os
```
gregbryan, line 6:
```suggestion
```
Duplicate.

run/MHD/2D/MHDCTOrszagTang/MHDCTOrszagTang__test_DivB_CT.py
```diff
@@ -3,7 +3,8 @@
 ###
 
 import os
-from yt.mods import *
+import os
```
gregbryan, line 6:
```suggestion
import os
```
Duplicate

run/MHD/2D/MHDCTOrszagTangAMR/MHDCTOrszagTangAMR__test_DivB_CT.py
```diff
@@ -3,7 +3,8 @@
 ###
 
 import os
-from yt.mods import *
+import os
```
gregbryan, line 6:
```suggestion
```
Duplicate

gregbryan:
> Now that the yt infrastructure we rely on is going away, we may want to think about moving to something like what we've setup for Enzo-E. That system can be parallelized automatically using circleci.

Britton -- Can you be more specific about what exactly is going away?

jwise77:
> Now that the yt infrastructure we rely on is going away, we may want to think about moving to something like what we've setup for Enzo-E. That system can be parallelized automatically using circleci.

That's good to know about the answer testing infrastructure going away. I've made issue #192 to remind us about it.

.circleci/config.yml
```diff
@@ -24,14 +28,21 @@ commands:
...
             source $HOME/venv/bin/activate
             pip install --upgrade pip
             pip install --upgrade wheel
             pip install --upgrade setuptools
-            pip install 'matplotlib<3.3.0'
-            pip install Cython numpy h5py fastcache flake8 nose yt==3.6.1 girder-client
+            pip install Cython numpy h5py fastcache flake8 nose girder-client matplotlib pytest gitpython # yt
+            # Temporary workaround to use updated answer testing from jwise's repository
+            # until next yt point release
+            git clone https://github.com/jwise77/yt
```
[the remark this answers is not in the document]
jwise77: My PR (yt-project/yt#3998) was merged earlier this week. I'll change this to the yt dev repository in a commit after my reply. We should switch back to pip after the next yt point release. I've made issue #193 to remind me.

brittonsmith:
> > Now that the yt infrastructure we rely on is going away, we may want to think about moving to something like what we've setup for Enzo-E. That system can be parallelized automatically using circleci.
> 
> Britton -- Can you be more specific about what exactly is going away?

The yt project is transitioning its answer testing from nose to pytest. The `AnswerTesting` class we import in `test_runner.py` from yt will be quite different from how it is now. I'm not sure what compatibility the new pytest framework will have with our current setup. I will investigate.

gregbryan:
This looks ready to go -- thanks @jwise77 for fixing testing; now we can address many outstanding pull requests.  I'm merging.

PR #197, opened 2022-08-22
Subtract previous B-field contribution

nirmitsakre:
Old file: Subtracted previous B-field contribution to Total Energy for DualE only and forgot to subtract for density <= 1 (else condition):

This file: Subtracted previous B-field contribution to Total Energy for all conditions.


gregbryan (approved):

pgrete (approved):
LGTM

PR #198, opened 2022-10-14
Smartstar

fearmayo:
Updates to the Active Particles code. Changes include porting across the supernova functionality that was part of the Star Particles machinery to the Smartstar flavour of Active Particles. I also pulled in Azton Wells changes that fixed the bug with supernovae but on accessing the bug fixes I don't think they are necessary as part of the Active Particles framework. In the AP framework the APs have their own grid and AP grids are always at the maximum refinement level. This should, unless I'm mistaken, negate the need for the bugfixes that are necessary in the Star Particle class as energy/metals/mass cannot be deposited into coarser grids as a result. 

Unfortunately since I pulled in Azton's changes early on into my branch this PR has a mix of his changes and mine. I'm not sure there is an easy way now to back his changes out? 

Actually looking through the file changes I think the easiest thing may be to just weed out the remaining changes that are specific to the Mechanical Stars (i.e. Azton's stuff) and then what is left will be specific to the Smartstars only.  The merge history will show all of this stuff but maybe that's not such a bug deal?

@azton @jwise77 @bwoshea any thoughts?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,7 +34,7 @@ commands:
             pip install --upgrade pip
             pip install --upgrade wheel
             pip install --upgrade setuptools
-            pip install Cython numpy h5py fastcache flake8 nose girder-client matplotlib pytest gitpython # yt
+            pip install Cython numpy h5py fastcache flake8 nose3 girder-client matplotlib pytest gitpython # yt
             # Temporary workaround to use updated answer testing from the yt dev repository
             # instead of pip until next yt point release
             git clone https://github.com/yt-project/yt
@@ -189,7 +189,7 @@ commands:
 jobs:
   test-suite:
     docker:
-      - image: cimg/python:3.8.5
+      - image: cimg/python:3.10
 
     resource_class: large
     working_directory: ~/enzo-dev
@@ -200,13 +200,13 @@ jobs:
 
       - restore_cache:
           name: "Restore dependencies cache."
-          key: dependencies-v3
+          key: dependencies-v5
 
       - install-dependencies
 
       - save_cache:
           name: "Save dependencies cache"
-          key: dependencies-v3
+          key: dependencies-v5
           paths:
             - ~/.cache/pip
             - ~/venv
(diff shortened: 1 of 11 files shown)


azton:
removing the Mechanical stars work sounds good to me.  I can (at some point when time allows) put in a pull request for that specific functionality.  The history will be more messy but the PR's will be more well defined and easy to follow.

fearmayo:
> removing the Mechanical stars work sounds good to me. I can (at some point when time allows) put in a pull request for that specific functionality. The history will be more messy but the PR's will be more well defined and easy to follow.

It's more I'm wondering about whether your fix for the deposition bug in the Star Particle class is needed in the Active Particle implementation. I don't think it is but I could be wrong. We should probably try to get some of your work pulled in in the medium term though especially since your work does for sure fix the bug in the Star Particle class

azton:
Ah.  I am not familiar with how the active particles couple to the grid, so I'm not too useful in this context.

bwoshea:
@fearmayo (and @jwise77 , @azton) I've agreed to take the lead on getting this PR across the finish line.  It seemed like there was some agreement in the discussion back in October about what to do with Azton's star particle deposition fixes - i.e., back those out and merge them in via a separate PR since they are not logically connected to this PR.  Given that it's a useful bugfix and Azton has moved to a new position, I think that it's better to just keep it in this PR and rename the PR so it also makes it clear that there's a bugfix.  What do people think?

bwoshea:
@fearmayo you're going to need to update the PR so the tests stop failing.  PR #210 and PR #211 fix a documentation issue with intersphinx and modify grackle's primary branch to be named `main` instead of `master` in the CI tests.  If you merge in all changes up to the current tip of enzo's `main` things should start working again.  Thank you!

fearmayo:
> @fearmayo you're going to need to update the PR so the tests stop failing. PR #210 and PR #211 fix a documentation issue with intersphinx and modify grackle's primary branch to be named `main` instead of `master` in the CI tests. If you merge in all changes up to the current tip of enzo's `main` things should start working again. Thank you!

@bwoshea This PR has been merged with the main branch so it's passing the tests now. As for Azton's changes I backed them out sometime ago and they are no longer part of this PR. I did initially merge with Azton's branch but I later found that unnecessarily complicated and backed them out. This PR is more or less solely focused on the Smartstar setup.

bwoshea:
Thanks, @fearmayo .  I'll get to work on this soon.  @jwise77 , are you also willing to review this PR in the next week or two?

jwise77:
Can do!

PR #191, opened 2022-07-05
Isolated Galaxy Sim Updates

clairekope:
Code additions used for [my recent paper ](https://arxiv.org/abs/2206.04706) that originally started as a fork from Devin Silvia back on Bitbucket (and is now lost). Documentation has been updated for the relevant parameters.

**GalaxySimulation Problem Type & DiskGravity External Potential**

- Multiple options for CGM structure, including rotation. One of these profiles requires the use of Grackle to calculate the cooling function on arbitrary data; this is encapsulated in GrackleCustomCoolRate
- Ability to initialize chemistry fields closer to equilibrium using precomputed pyGrackle tables. A sample script is included in the input directory. As noted in the docs, copying this feature to a different problem type may require manually overriding mu in ComputeTemperatureField, or the initial temperature will not be correctly calculated when saved to disk.
- DiskGravity stellar potentials (disk & bulge) are now accounted for when calculating the disk's circular velocity
- The DiskGravity external potential now includes an NFW potential (PointSourceGravity and DiskGravity are mutually exclusive within the isolated galaxy problem type)
- If using the DiskGravity potential, NFW halo parameters (mass & concentration) are automatically copied over to the GalaxySimulation counterparts to ensure consistency

**Star Particle Creation & Feedback**
- Added self-bound criterion and ability to use H2 mass to creation method 0 (Cen & Ostriker) as found in Hopkins et al. 2013
- Added parameters for toggling individual star formation criteria in creation method 0
- Can adjust temperature threshold used in creation method 0
- Can ramp the efficiency of thermal feedback a la minimum particle mass

PR #201, opened 2023-03-07
Resurrect GalaxySim Functionality

clairekope:
Early in the development of merge #191 I had deleted some code I didn't need, forgotten about it, and now I've put it back for full backwards compatibility (as it were) of Enzo and the GalaxySimulation problem type. To reconcile the old code with my changes I have added some new parameters.

Primarily, these parameters decide: which static dark matter potential to use (both during disk initialization and during sim runtime), and whether to use an isothermal initial disk or the thermally pressure balanced disk from Tonnesen & Bryan 09.

Bonus content: 
* there's a machine file for SDSC Expanse
* CircleCI will save `out.compile` as an artifact of `test-compile-options`, so that if linking fails, it can be debugged.


clairekope:
test-compile-options is failing while linking but doesn't fail locally.
The test-suite is failing to compile `RadiativeTransferWriteParameters.C` which was not modified in this PR. Error is about `size_t` being undefined, with suggested alternative `__size_t`. This error doesn't occur when building locally.

bwoshea (approved):
Looks good to me!

bwoshea:
Looks good.  @clairekope there's a merge conflict - if you can fix that we can go ahead and merge!

clairekope:
> Looks good. @clairekope there's a merge conflict - if you can fix that we can go ahead and merge!

Done!

clairekope:
Doc build is fixed by #207 and works offline

bwoshea:
Since the docs-build issue is fixed in another PR, I am going to go ahead and merge this PR.

PR #202, opened 2023-03-18
Adding updated H2 self-shielding fit from Wolcott-Green & Haiman 2019

mihirskulkarni:
Currently, RadiationShield = 2 selects H2 self-shielding from Wolcott-Green et al. (2011). This fit was updated in Wolcott-Green & Haiman (2019) for a wider range of density and temperatures. I have added the new self-shielding fit for RadiationShield = 3 and kept the old fit with RadiationShield = 2 for backward compatibility, but the new fit should be used by default if one wants to include H2 self-shielding.


gregbryan (approved):
Looks good to me -- thanks Mihir!  The docs failure is unrelated (and is dealt with in PR#206).

PR #205, opened 2023-04-26
Add Make.mach.conda

matthewturk:
It can be convenient to install using a conda environment.  This adds a makefile that uses a conda environment for its packages.

In some specific cases, this becomes extra useful for using embedded python via libyt, which will be addressed in a forthcoming PR.


gregbryan (approved):
Thanks Matt!

jwise77 (approved):

jwise77:
I'm merging even though it's failing the docs build, which was fixed in #207.

PR #206, opened 2023-05-01
removed intersphinx examples in deprecated format

gregbryan:
Quick fix for failed docs build (removed what seems like unused intersphinx examples that use deprecated format).


gregbryan:
Superseded by PR#207

PR #207, opened 2023-05-03
Sphinx update

clairekope:
Should fix CircleCI "warning as error" failure for documentation build. Addresses a deprecation warning for Sphinx 7.

PR #210, opened 2023-05-08
Updating intersphinx syntax and readthedocs config

jwise77:
Intersphinx isn't supported in sphinx v7.0, so I'm removing it. The same was done by @mabruzzo in Enzo-E (enzo-project/enzo-e#305).

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,7 @@
+version: 2
+
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.11"
+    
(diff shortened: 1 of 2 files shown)


mabruzzo (approved):
Looks good to me.

The read-the-docs test is failing due to an unrelated issue (with the message ``Could not import extension sphinx.builders.linkcheck (exception: urllib3 v2.0 only supports OpenSSL 1.1.1+, currently the 'ssl' module is compiled with OpenSSL 1.0.2n  7 Dec 2017. See: https://github.com/urllib3/urllib3/issues/2168)``).

We currently see this failure in some Grackle PRs. I was starting to look into resolving this.

jwise77:
Thanks I'll address the ssl issue and then we can merge

jwise77:
Also, I see that this fix was included in #207. But I've updated the URLs to point to the current ones.

PR #216, opened 2023-06-30
Fix GalaxySimulation boundary to work correctly with DualEnergyFormalism

gregbryan:
This small PR makes a few corrections to the specialized ExternalBoundary machinery that is part of the GalaxySimulation framework so that the gas coming in the boundary correctly respects the DualEnergyFormalism, HydroMethod, and Multispecies flags.

### DIFF
--- a/src/enzo/ExternalBoundary_SetGalaxySimulationBoundary.C
+++ b/src/enzo/ExternalBoundary_SetGalaxySimulationBoundary.C
@@ -60,6 +60,20 @@ int ExternalBoundary::SetGalaxySimulationBoundary(FLOAT time)
     ENZO_FAIL("Error in IdentifyPhysicalQuantities.\n");
   }
 
+  int DeNum = FindField(ElectronDensity,BoundaryFieldType,NumberOfBaryonFields);
+  int HINum = FindField(HIDensity,BoundaryFieldType,NumberOfBaryonFields);
+  int HIINum = FindField(HIIDensity,BoundaryFieldType,NumberOfBaryonFields);
+  int HeINum = FindField(HeIDensity,BoundaryFieldType,NumberOfBaryonFields);
+  int HeIINum = FindField(HeIIDensity,BoundaryFieldType,NumberOfBaryonFields);
+  int HeIIINum = FindField(HeIIIDensity,BoundaryFieldType,NumberOfBaryonFields);
+  int HMNum = FindField(HMDensity,BoundaryFieldType,NumberOfBaryonFields);
+  int H2INum = FindField(H2IDensity,BoundaryFieldType,NumberOfBaryonFields);
+  int H2IINum = FindField(H2IIDensity,BoundaryFieldType,NumberOfBaryonFields);
+  int DINum = FindField(DIDensity,BoundaryFieldType,NumberOfBaryonFields);
+  int DIINum = FindField(DIIDensity,BoundaryFieldType,NumberOfBaryonFields);
+  int HDINum = FindField(HDIDensity,BoundaryFieldType,NumberOfBaryonFields);
+  int MultiSpecies = (HINum == -1) ? 0 : 1;
+
 	/* Determine if we're using metallicity (as a color field) */
 	int MetalNum = FindField(Metallicity,BoundaryFieldType,NumberOfBaryonFields);
 	int UseMetallicityField = (MetalNum == -1) ? 0 : 1;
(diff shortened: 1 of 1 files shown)


src/enzo/ExternalBoundary_SetGalaxySimulationBoundary.C
```diff
@@ -60,6 +60,20 @@ int ExternalBoundary::SetGalaxySimulationBoundary(FLOAT time)
...
+  int HeINum = FindField(HeIDensity,BoundaryFieldType,NumberOfBaryonFields);
+  int HeIINum = FindField(HeIIDensity,BoundaryFieldType,NumberOfBaryonFields);
+  int HeIIINum = FindField(HeIIIDensity,BoundaryFieldType,NumberOfBaryonFields);
+  int HMNum = FindField(HMDensity,BoundaryFieldType,NumberOfBaryonFields);
+  int H2INum = FindField(H2IDensity,BoundaryFieldType,NumberOfBaryonFields);
+  int H2IINum = FindField(H2IIDensity,BoundaryFieldType,NumberOfBaryonFields);
+  int DINum = FindField(DIDensity,BoundaryFieldType,NumberOfBaryonFields);
+  int DIINum = FindField(DIIDensity,BoundaryFieldType,NumberOfBaryonFields);
+  int HDINum = FindField(HDIDensity,BoundaryFieldType,NumberOfBaryonFields);
+  int MultiSpecies = (HINum == -1) ? 0 : 1;
```
clairekope, line 75: I don't believe setting `MultiSpecies` in this way will allow for `MultiSpecies > 1` (line 332) to ever occur. Can `MultiSpecies` not just be set from the parameter?

src/enzo/ExternalBoundary_SetGalaxySimulationBoundary.C
```diff
@@ -269,8 +322,26 @@ int ExternalBoundary::SetGalaxySimulationBoundary(FLOAT time)
...
+                if (MultiSpecies > 0)  {
+                  BoundaryValue[HIINum][dim][0][index] = 0.5 * CoolData.HydrogenFractionByMass * BoundaryValue[DensNum][dim][0][index];
+                  BoundaryValue[HeIINum][dim][0][index] = 0.2 * BoundaryValue[DensNum][dim][0][index] * (1.0 - CoolData.HydrogenFractionByMass); //get rid of factor of 4
+                  BoundaryValue[HeIIINum][dim][0][index] = 0.2 * BoundaryValue[DensNum][dim][0][index] * (1.0 - CoolData.HydrogenFractionByMass); //get rid of factor of 4
+                  BoundaryValue[HeINum][dim][0][index] = (1.0-CoolData.HydrogenFractionByMass)*BoundaryValue[DensNum][dim][0][index] - BoundaryValue[HeIINum][dim][0][index] - BoundaryValue[HeIIINum][dim][0][index];
+                  BoundaryValue[HINum][dim][0][index] = CoolData.HydrogenFractionByMass * BoundaryValue[DensNum][dim][0][index] - BoundaryValue[HIINum][dim][0][index];
+                  BoundaryValue[DeNum][dim][0][index] = BoundaryValue[HIINum][dim][0][index] + 0.25*BoundaryValue[HeIINum][dim][0][index] + 0.5*BoundaryValue[HeIIINum][dim][0][index];
+                  if (MultiSpecies > 1) {
+                    BoundaryValue[HMNum][dim][0][index] =1e-10;
```
clairekope, line 333: Is it possible to not fix these values in order to prevent unexpected behavior later? Or is this the standard "initial value" for these species.

gregbryan: We have to set the boundary values to something (otherwise they could be undefined and corrupt the solution). Properly, these should be controlled with a parameter, but right now there is no way to know their true value. This code sets them to something vaguely appropriate -- generally, the wind flowing in is hot, so the ionization fractions will rapidly adjust to their correct equilibrium value, so in practice, the exact values here are not essential (but they need to be set). I will add a comment nothing this.

clairekope: I think leaving a comment explaining that the values should rapidly equilibrate is a good solution.

clairekope (approved):

src/enzo/ExternalBoundary_SetGalaxySimulationBoundary.C
```diff
@@ -268,9 +320,27 @@ int ExternalBoundary::SetGalaxySimulationBoundary(FLOAT time)
...
+		  }
+                  if (MultiSpecies > 2) {
+                    BoundaryValue[DINum][dim][0][index] = 1e-10;
+                    BoundaryValue[DIINum][dim][0][index] = 1e-10;
+                    BoundaryValue[HDINum][dim][0][index] = 1e-10;
+		  }
+		}
 		if( UseMetallicityField )
-			BoundaryValue[MetalNum][dim][0][index] = 1.0e-10;
+		  BoundaryValue[MetalNum][dim][0][index] = 0.1 * BoundaryValue[DensNum][dim][0][index] * CoolData.SolarMetalFractionByMass; // 0.1 solar metallicity
```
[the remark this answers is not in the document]
gregbryan: No good justification -- a parameter needs to be created here but I feel it is a bit beyond the scope of this PR (this line is just to make a slightly less egregious choice). I propose that we accept this PR as is and I create an issue for this problem (as described below).

New issue:
A parameter which controls the metallicity of the inflowing gas in this problem should be created and connected (requires changes to GalaxySimulationInitialize.C, Grid_GalaxySimulationInitialize.C, and ExternalBoundary_SetGalaxySimulationBoundary). In particular, a new global parameter in the GalaxySimulationRPSWind family would need to be created, set and used in those routines. Currently the metallicity inflow is hard-wired to 0.1 solar as the least bad thing.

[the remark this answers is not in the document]
bwoshea: That sounds great to me.  I'll approve and merge.

bwoshea (approved):

src/enzo/ExternalBoundary_SetGalaxySimulationBoundary.C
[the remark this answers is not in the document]
clairekope: @gregbryan I went ahead and made this as issue #236

PR #221, opened 2023-09-20
Added StarFormationOncePerRootGridTimeStep to Normal Star

clairekope:
Extended existing `StarFormationOncePerRootGridTimeStep` functionality to Normal Star (star_maker2.F). Also adjusted what grids `StarFormationOncePerRootGridTimeStep` triggers on. Previously, it only allowed star formation on grids that were at the maximum refinement level. Since each star maker routine checks that a cell is the most refined representation for its volume, this restriction is unnecessary and prevents star formation in cells where it would otherwise occur.


bwoshea (approved):
This looks good to me!

brittonsmith (commented):
The code looks good to me. @clairekope, could you also update the documentation on this parameter? It's in `doc/manual/source/parameters/index.rst`.

brittonsmith (approved):
Thanks for updating the docs! Looks great to me.

bwoshea:
Thanks, @brittonsmith !  I'm going to merge the PR now.

Issue #222, opened 2023-09-20
CircleCI Python Too Old


clairekope:
CircleCI is failing to install yt because of unresolvable dependencies around NumPy. The requested NumPy version (>=1.25, <2.0) is available for Python >= 3.9, but not whatever version CircleCI is defaulting to.

PR #223, opened 2023-09-20
Update CircleCI Python several versions for future proofing

clairekope:
Addresses #222

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -240,7 +240,7 @@ jobs:
 
   test-compile-options:
     docker:
-      - image: cimg/python:3.11.5
+      - image: cimg/python:3.10.13
 
     resource_class: large
     working_directory: ~/enzo-dev
@@ -271,7 +271,7 @@ jobs:
 
   docs-build:
     docker:
-      - image: cimg/python:3.11.5
+      - image: cimg/python:3.10.13
 
     working_directory: ~/enzo-dev
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -217,7 +217,7 @@ jobs:
 
       - restore_cache:
           name: "Restore test answers."
-          key: test-answers-v16b
+          key: test-answers-v16a
 
       - build-and-test:
           tag: gold-standard-v16
@@ -226,7 +226,7 @@ jobs:
 
       - save_cache:
           name: "Save test answers cache."
-          key: test-answers-v16b
+          key: test-answers-v16a
           paths:
             - ~/enzo_test/push_suite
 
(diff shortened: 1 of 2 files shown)


clairekope:
That's certainly an easier flag. Though for some reason it's not being passed when `ffte4X.F90` is getting compiled...

gregbryan:
The problem is that the answer testing works by compiling a version with the current gold-standard (which does not have this flag) and then compiling a version with this PR changes. The PR changes are clearly working because the test-compile-options test passes, but the first step of the test-suite (compilation of the current gold-standard) is the thing that is failing. There is no way to fix that in a PR except by updating the gold-standard.  I think that is probably what we need to do, go from gold-standard-15 to gold-standard-16. Unfortunately, this takes a few steps which are outlined here: https://github.com/enzo-project/enzo-e/issues/307

clairekope:
I did not realize that the gold standard version was also compiled. I'll create a new issue for tracking the gold standard update.

clairekope:
I've pushed a lightweight tag as part of this PR that should the gold standard to building with `-std=legacy` (at least for the ubuntu machine file)

gregbryan:
Thanks Claire!  Now it looks like it is failing due to a pynose problem. I'm guessing that you may need to bump dependencies-v3 to dependencies-v4 in config.yml to clear the dependencies cache, since you bumped the CircleCI image.

clairekope:
Tried Python 3.10.13, but nose might still not work (https://github.com/xlcnd/isbnlib/issues/95) despite the issues of not(?) having python3 in the circleci image?

clairekope:
Path forward: return to the 3.11.5 image and migrate from nose to pytest. Nose hasn't been updated since 2015 anyway.

clairekope:
Tests passed but there are unyt errors in the reference tests

clairekope:
12 of the reference tests fail due to a change with unyt v3. This is fixed by a PR on yt: https://github.com/yt-project/yt/pull/4764

There is also an issue that, when comparing the new version of Enzo to the standard, the standard's answers don't seem to be saved.

clairekope:
I have figured out the nuances of CircleCI. For future reference: 
* when updating software versions for the test toolchain, make sure to update the dependency cache or you will get a vague error about a missing Python executable
* when updating the gold standard, first update the CircleCI config to use this new standard. This means updating the reference/answer tag as well as the name for the answer cache. Commit these changes and attach a tag to this same commit. Placing a tag on an older commit than the one where the CircleCI config was updated will mean CircleCI will not cache answers correctly.

Please please please squash these commits when merging.

gregbryan (approved):
Looks good to me -- thanks Claire for persevering though all of that!

bwoshea:
Looks good to me!

bwoshea (approved):
Looks great, I'm approving and merging.

Issue #200, opened 2023-02-24
FuzzyDarkMatter test runs not working properly


drenniks:
A few issues with the FuzzyDarkMatter test run (calling @jwise77 ): 

- FDMCosmology.enzo and FDMTest.enzo are saved as executables 
- FDMCollapse.enzo runs but doesn't spit out any datasets (except for the first DD). 
- FDMCosmology.enzo breaks with the following error: 

```
MPI_Init: NumberOfProcessors = 1
warning: the following parameter line was not interpreted:
GreensFunctionMaxNumber = 100
WARNING! PartitionNestedGrids = 1 forces Parallel IO = 1
Global Dir set to .
Initialdt in ReadParameterFile = 0.000000e+00
InitializeNew: Starting problem initialization.
CosmologySimulation: check CellFlaggingMethod.
magnetic field: dim 0, 0.003622 0.000000e+00 
magnetic field: dim 1, 0.003622 0.000000e+00 
magnetic field: dim 2, 0.003622 0.000000e+00 
Start loop creating initial grid hierarchy entries, from one
Start loop initializing generated grids, from zero
RH: CosmologySimulation: Initializing grid 0
Call CSIG for gridnum 0 with TR 1 and Dname GridDensity
HDF5-DIAG: Error detected in HDF5 (1.10.2) thread 140450751264576:
  #000: H5Dio.c line 160 in H5Dread(): selection+offset not within extent
    major: Dataspace
    minor: Out of range
Caught fatal exception:

   'IO Problem'
at ReadFile.C:292
```

I haven't had the time to dig into these issues further, these are just the first problems I saw with them.

FDMTest.enzo **does** work.

PR #244, opened 2025-03-11
Pull HydrogenFractionByMass value back from Grackle to Enzo.

brittonsmith:
This fixes a bug introduced by [Grackle PR #200](https://github.com/grackle-project/grackle/pull/200). This PR changed the way HydrogenFractionByMass takes its default value, first setting it to FLOAT_UNDEFINED, detecting that, and then setting it to 0.76. This leaves Enzo's internal version of this parameter set to FLOAT_UNDEFINED which has catastrophic consequences almost immediately within Enzo. This fixes that by pulling Grackle's version of this parameter back to Enzo so they are in sync.

### DIFF
--- a/src/enzo/GrackleReadParameters.C
+++ b/src/enzo/GrackleReadParameters.C
@@ -243,6 +243,13 @@ int GrackleReadParameters(FILE *fptr, FLOAT InitTime)
     ENZO_FAIL("Error in Grackle initialize_chemistry_data.\n");
   }
 
+  /*
+    Return this value to Enzo's storage as Grackle will change this value during
+    initialization (see Grackle PR #200
+    https://github.com/grackle-project/grackle/pull/200).
+  */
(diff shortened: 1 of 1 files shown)


jwise77 (approved):

gregbryan (approved):

gregbryan:
The docs build is breaking because I think read-the-docs now needs the following to be added to .readthedocs.yaml

```
 sphinx:
   # Path to your Sphinx configuration file.
   configuration: doc/source/conf.py
```

brittonsmith:
I've had a look at the failing test. It is PhotonTestMultiFrequency. What's happening is that the test answers are no longer cached and so it is failing trying to generate the gold standard. This PR returns this test to a passing state, so I think that means we need to merge this and update the gold standard.

brittonsmith:
@gregbryan, @jwise77 is this good to go?

jwise77:
Yep, merging!

PR #245, opened 2025-03-12
Return metal density cap.

brittonsmith:
This fixes an issue introduced in [Grackle PR #121](https://github.com/grackle-project/grackle/pull/121), where we removed a seemingly arbitrary cap on the metal density. This cap is necessary in Enzo as star formation feedback can produce grid cells where the metal density exceeds the gas density. This may not be either the best fix or the best place to put it, but this PR is as close to restoring prior functionality as is possible without putting it back into Grackle, which we don't want to do. I'm happy for this PR to be a place for a broader conversation.


gregbryan (approved):

bwoshea (approved):

brittonsmith:
The tests will pass here once PR #244 has been merged.

PR #246, opened 2025-04-03
Fix bugs in H2 self-shielding implementations

thinhhn2:
---
name: Thinh Huu Nguyen
about: Fix bugs in $H_2$ self-shielding implementations
---

**Pull request summary**

Fix bugs in the Sobolev-like approximation in the $H_2$ self-shielding implementation.

**Describe your pull request**

There are two bugs in the Sobolev-like approximation formulas of the $H_2$ self-shielding implementations. The Sobolev-like approximation is activated by setting _RadiativeTransferOpticallyThinH2 == 1_ and _RadiationShield == 2_ in the parameter file. The code to implement is _[**solve_rate_cool.F**](https://github.com/enzo-project/enzo-dev/blob/main/src/enzo/solve_rate_cool.F#L1209)_. The two bugs are:

1. On Line [1209](https://github.com/enzo-project/enzo-dev/blob/main/src/enzo/solve_rate_cool.F#L1209), we compute the variable $b_5$ (or $b_{\text{Doppler}}$) to get the self-shielding factor using the fit in Equation 12 of [Wolcott-Green et al 2011](https://ui.adsabs.harvard.edu/abs/2011MNRAS.418..838W/abstract). Since the equation concerns the Doppler line broadening of $H_{2}$'s absorption band, the mass of $H_{2}$ should be used in stead of the mass of atomic hydrogen in this line (i.e., `2.0_RKIND * mass_h` instead of just `mass_h`).

2. On Line [1201](https://github.com/enzo-project/enzo-dev/blob/main/src/enzo/solve_rate_cool.F#L1201), the Sobolev-like length is calculated as $L_{Sob} = \rho/\nabla\rho$, following Equation 4 of [Gnedin et al 2009](https://ui.adsabs.harvard.edu/abs/2009ApJ...697...55G/abstract). However, in [Gnedin and Kravtsov 2011](https://ui.adsabs.harvard.edu/abs/2011ApJ...728...88G/abstract), the same author said that the equation in Gnedin et al 2009 had a typo and that a factor of 2 was missing in the denominator, and the correct equation should be $L_{Sob} = \rho/(2\nabla\rho)$ (Equation A10). Later work of this author [Gnedin and Draine 2014](https://ui.adsabs.harvard.edu/abs/2014ApJ...795...37G/abstract) also used $L_{Sob} = \rho/(2\nabla\rho)$ in their analysis. Therefore, a factor of 2 should be added to the formula on line 1201.

### DIFF
--- a/src/enzo/solve_rate_cool.F
+++ b/src/enzo/solve_rate_cool.F
@@ -1198,7 +1198,7 @@ c     Exclude directions with (drho/ds > 0)
                   endif
                enddo
 c     (rho / divrho) is the Sobolev-like length in cell widths
-               l_sobolev = min(dx_cgs * d(i,j,k) / abs(divrho), xbase1)
+               l_sobolev = min(dx_cgs * d(i,j,k) / (2.0_RKIND*abs(divrho)), xbase1)
                N_H2 = dom*H2I(i,j,k) * l_sobolev
 c               if (N_H2.gt.1e14 .and. divrho.gt.100*tiny) then
 c                  write(6,*) "test"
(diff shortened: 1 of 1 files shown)


jwise77 (approved):

jwise77:
I'm approving with one review since this is a minor change and has already been accepted into Grackle.

PR #249, opened 2025-08-23
Fix and enhancement for libyt in situ feature

cindytsai:
## Fix 

- Set unit for `DarkMatter` particle type
- Enzo `particle_mass` is in fact _mass density_, so we need to convert to real mass before setting the pointer.
- `precision-32/64` and `particles-32/64` can compile

## Enhancement

- Add `dtLibytCall` and `TimeLastLibytCall` to govern the time-based libyt call routine in Enzo.
- Check if we need to call libyt routine at initialization steps.

### DIFF
--- a/src/enzo/Grid_ConvertToLibyt.C
+++ b/src/enzo/Grid_ConvertToLibyt.C
@@ -52,12 +52,12 @@ void grid::ConvertToLibyt(int LocalGridID, int GlobalGridID, int ParentID, int l
      * In here, we'll happily do whatever we're told.
      *
      * */
-    float grid_unit_length[MAX_DIMENSION] = {1.0, 1.0, 1.0};
+    float cell_width[MAX_DIMENSION] = {1.0, 1.0, 1.0};
     for (int i = 0; i < MAX_DIMENSION; i++) {
         GridInfo.grid_dimensions[i] = (this->GridEndIndex[i]) - (this->GridStartIndex[i]) + 1; // this is active dimension
         GridInfo.left_edge[i] = this->GridLeftEdge[i];
         GridInfo.right_edge[i] = this->GridRightEdge[i];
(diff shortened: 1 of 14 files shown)


gregbryan (approved):
Looks good to me

cindytsai:
Would someone also be able to help review this PR? Thanks!

src/enzo/Grid_ConvertToLibyt.C
```diff
@@ -52,11 +52,12 @@ void grid::ConvertToLibyt(int LocalGridID, int GlobalGridID, int ParentID, int l
      * In here, we'll happily do whatever we're told.
      *
      * */
-  
+    float grid_unit_length[MAX_DIMENSION] = {1.0, 1.0, 1.0};
     for (int i = 0; i < MAX_DIMENSION; i++) {
         GridInfo.grid_dimensions[i] = (this->GridEndIndex[i]) - (this->GridStartIndex[i]) + 1; // this is active dimension
         GridInfo.left_edge[i] = this->GridLeftEdge[i];
         GridInfo.right_edge[i] = this->GridRightEdge[i];
+        grid_unit_length[i] = (GridInfo.right_edge[i] - GridInfo.left_edge[i]) / GridInfo.grid_dimensions[i];
```
jwise77: Just confirming that you want to change this value from 1.0 to the dx of this grid. I believe you do because of the product below with the code particle mass (density). I'd recommend renaming it to indicate that it's the cell width and not the grid unit length.

jwise77 (changes requested):
All looks good with the exception of my minor comment. Thanks for contributing!

src/enzo/Grid_ConvertToLibyt.C
cindytsai: Thanks for the suggestion!

jwise77 (approved):
Thanks!

PR #253, opened 2026-08-03
Subgrid autoadjust floor - 

tumlinson:
# Make the SubgridSizeAutoAdjust floor a parameter (SubgridSizeAutoAdjustMinimum)

## Summary

By default `DetermineSubgridSizeExtrema.C` estimates a maximum subgrid size from the level's cell count and the processor count: 

```
MaximumSubgridSize = NumberOfCells / (NumberOfProcessors * OptimalSubgridsPerProcessor)
```

and then clamps it from below with a hard-coded `#define MINIMUM_SIZE 2000` (line 35).  With large processor counts and deep refinement the estimate this fails to capture the fact that a relatively small number of grids exist on the highest level, with only a few per process, leading to poor load balancing. 

This PR makes the floor settable as **`SubgridSizeAutoAdjustMinimum`**, defaulting to 2000 so existing behavior is unchanged. The effect is to create more grids at the highest level, enabling better load balancing. In practice this parameter can remain unset (and so defaulting to 2000) or set at a lower value with some experimentation to achieve a good balance. Testing indicates that using SubgridSizeAutoAdjustMinimum = 512 is a ~30% speedup on production-style FOGGIE runs (nref11 natural runs on Tempest). This is an effect of the smaller more broadly distributed grids. 

## Motivation

On a deeply refined level with ~10^6 cells at 128 processors, the current estimate asks for a few hundred cells per subgrid and is forced to 2000 by the hardcoded floor. The consequence is a shortage of subgrids to distribute:

| level | subgrids | processors | subgrids per processor |
|---|---|---|---|
| 8 | 456 | 128 | 3.6 |
| 9 | 222 | 128 | 1.7 |
| 10 | 195 | 128 | **1.5** |

With 1.5 subgrids per processor and sizes on that level spanning 24 to 15,552 cells, the largest single subgrid holds **2.35x an even share of the level's work**. A subgrid cannot be split across processors, so no load balancer can compensate; the imbalance is structural. Measured per-subcycle imbalance on the test configuration was 2.71x max/mean, consistent with the 2.35x.

Lowering the floor lets the existing estimate apply, producing more and smaller subgrids and a correspondingly better balance. The trade-off is more ghost-zone and boundary-exchange work per cell, so there is an optimum rather than a monotonic gain — see below.

## Change

Functionally one line, in `DetermineSubgridSizeExtrema.C`:

```diff
-  MaximumSubgridSize = max(MaximumSubgridSize, MINIMUM_SIZE);
+  MaximumSubgridSize = max(MaximumSubgridSize,
+                           (long_int) SubgridSizeAutoAdjustMinimum);
```

plus the usual parameter plumbing (declaration in `global_data.h`, default in
`SetDefaultGlobalValues.C`, read/write in `ReadParameterFile.C` /
`WriteParameterFile.C`) and documentation in
`doc/manual/source/parameters/index.rst` and
`doc/manual/source/user_guide/RunningLargeSimulations.rst`.

`MINIMUM_SIZE` is retained and still supplies the default, so this reads as "make an existing constant configurable" rather than a behaviour change.

The parameter only has effect when `SubgridSizeAutoAdjust = 1` (the default) *and* the level's estimate falls below the floor, so it is inert on small or shallow runs.


tumlinson:
This PR was mistakenly filed against enzo-dev instead of enzo-foggie fork. Sorry!

